PR #16, opened 2016-02-11
for string regex questions, translate ignore_case flag to pattern mod…

githubapprentice:
…ifier
  I'm guessing this flag was intended to be ignore_case
  Looks like ignore_case is always set, so this uses its value rather than isset(ignore_case)


drlippman:
Impressive - you caught my error before I even made the code live :)

githubapprentice:
What's impressive is the thousands of lines of code you've written and shared.  This pull request was purely selfish.

PR #18, opened 2016-02-11
Add regex string to help

githubapprentice:
This adds explanation of the new regex flag for string-questions.
Happy to remove link to regex video if the python stuff is distracting


drlippman:
Thanks for updating the help file :)  Yeah, the braces as pattern delimiters won't work. I'll get the delimiter issue fixed up.

PR #19, opened 2016-02-27
Force attachment links in the last answer to open in…

githubapprentice:
… a new window to prevent loss of unsaved feedback or manual scores.

If while grading an essay question, an instructor clicks an attachment link to view a student's attachment, open the link in a new window/tab so any unsaved feedback or manual scoring are not lost.

### DIFF
--- a/assessment/displayq2.php
+++ b/assessment/displayq2.php
@@ -2031,7 +2031,8 @@ function makeanswerbox($anstype, $qn, $la, $options,$multi,$colorbox='') {
 					if (strpos($m[0],'target=')===false) {
 						$ret = '<a target="_blank" '.substr($m[0], 2);
 					} else {
-						$ret = $m[0];
+						//force links in the last answer (from attach.js) to open in a new window to prevent manual scores and feedback from being lost
+						$ret = preg_replace("/target=\"_self/","target=\"_new",$m[0]);
 					}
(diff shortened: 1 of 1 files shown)


githubapprentice:
Please help me double-check that this doesn't have adverse effects on other links that might be contained in the last answer or other situations where this code is used.

drlippman:
I don't foresee any issues with this.  However, you could accomplish the replacement more simply with str_replace than preg_replace, since you're not really using a regular expression.

It would probably be good to totally remove the "target" option from the attach tool.  I can't imagine any good reason to ever have an attachment open in the same window.

githubapprentice:
Removing the target option in attach.js would be great for this issue and I'd be happy to submit it.  Would that make upgrading/maintaining that plugin more complicated?

The motivation for this pull-request is I was grading an essay question and clicked on an a student's spreadsheet attachment.  It opened in the same window and after navigating back, my feedback and grade changes were lost.

drlippman:
I wrote that plugin specifically for imathas because using the built in link tool to attach files wasn't obvious to people.

If you want to make the changes to the attach plugin to remove the target option and always use _blank, go for it :)

drlippman:
A made a small tweak to change all targets to _blank, just to be safe.

PR #20, opened 2016-02-28
remove option to open attachment links in same window (minimal change)

githubapprentice:
In TinyMCE attach plugin, removes target option for opening link in the same window.

This version of the patch doesn't remove the code used to give the user option of setting the target.


githubapprentice:
I hope to submit another version of the patch that removes more code in case that's preferable.

drlippman:
Going with the total removal option.

PR #21, opened 2016-02-28
remove code in attach editor pluging related to setting the target

githubapprentice:
always use new window (_blank)


githubapprentice:
This pull-request is an alternative to a previous pull-request.  Close what's not needed.

Please check attach.js line 6.  I'm not actually sure what the templates variable does.

editor/plugins/attach/js/attach.js
```diff
@@ -3,7 +3,7 @@
 tinyMCEPopup.requireLangPack();
 
 var templates = {
-	"window.open" : "window.open('${url}','${target}','${options}')"
```
githubapprentice: Not sure what this does.

drlippman: Me neither.

PR #23, opened 2016-03-09
Allow categorizing questions by other assessments in which that question appears

githubapprentice:
e.g. a review (or diagnostic) assessment can guide students to a video lesson assessment based on the student's score on a question that's in both the current assessment and another assessment.  In that case, a link can be included in the Categorized Score Breakdown.

Happy to make changes if there's a better way to implement it or any unforeseen issues.

### DIFF
--- a/assessment/catscores.php
+++ b/assessment/catscores.php
@@ -86,6 +86,13 @@ function catscores($quests,$scores,$defptsposs,$defoutcome=0,$cid) {
 	foreach (array_keys($catscore) as $category) {
 		if (is_numeric($category)) {
 			continue;
+		} elseif (0==strncmp($category,"AID-",4)) { //category is another assessment
+			$categoryaid=substr($category,4);
+			$query = "SELECT name FROM imas_assessments WHERE id='$categoryaid' LIMIT 1";
+			$result = mysql_query($query) or die("Query failed : " . mysql_error());
+			$assessmentname = mysql_result($result, 0, 0);
+			//link to the other assessment
+			$catname="<a href='../assessment/showtest.php?cid=$cid&aid=$categoryaid' >$assessmentname</a>";
 		} else {
 			$catname = $category;
 		}
--- a/assessment/catscores.php
+++ b/assessment/catscores.php
@@ -88,7 +88,7 @@ function catscores($quests,$scores,$defptsposs,$defoutcome=0,$cid) {
 			continue;
 		} elseif (0==strncmp($category,"AID-",4)) { //category is another assessment
 			$categoryaid=substr($category,4);
-			$query = "SELECT name FROM imas_assessments WHERE id='$categoryaid' LIMIT 1";
+			$query = "SELECT name FROM imas_assessments WHERE id='$categoryaid' AND courseid='$cid' LIMIT 1";
 			$result = mysql_query($query) or die("Query failed : " . mysql_error());
 			$assessmentname = mysql_result($result, 0, 0);
 			//link to the other assessment
--- a/assessment/catscores.php
+++ b/assessment/catscores.php
@@ -87,7 +87,7 @@ function catscores($quests,$scores,$defptsposs,$defoutcome=0,$cid) {
 		if (is_numeric($category)) {
 			continue;
 		} elseif (0==strncmp($category,"AID-",4)) { //category is another assessment
-			$categoryaid=substr($category,4);
+			$categoryaid=intval(substr($category,4));
 			$query = "SELECT name FROM imas_assessments WHERE id='$categoryaid' AND courseid='$cid' LIMIT 1";
 			$result = mysql_query($query) or die("Query failed : " . mysql_error());
 			$assessmentname = mysql_result($result, 0, 0);
--- a/course/categorize.php
+++ b/course/categorize.php
@@ -96,16 +96,42 @@ END;
 		$questionlibs[$row[0]][] = $row[1];
 		$libnames[$row[1]] = $row[2];
 	}
-	$query = "SELECT iq.id,iq.category,iqs.description FROM imas_questions AS iq,imas_questionset as iqs";
+
+	//add assessment names as options
+	//find the names of assessments these questionsetids appear in
+	$query = "SELECT imas_questions.id AS qid,imas_questions.questionsetid AS qsetid,imas_assessments.id AS aid,imas_assessments.name ";
+	$query .= "FROM imas_questions, imas_assessments ";
+	$query .= "WHERE imas_questions.assessmentid=imas_assessments.id ";
+	$query .= "AND imas_questions.questionsetid = ANY (SELECT imas_questions.questionsetid FROM imas_questions WHERE imas_questions.assessmentid='$aid') ";
+	$query .= "AND imas_assessments.courseid='$cid' ";
+	$query .= "ORDER BY aid";
+	$result = mysql_query($query) or die("Query failed : " . mysql_error());
+	$assessmentnames = array();
+	$qsetidassessment = array();
+	while ($row = mysql_fetch_array($result,MYSQL_ASSOC)) {
+		//store the relevent assessment names
+		$assessmentnames[$row['aid']] = $row['name'];
+		if ($row['aid']!=$aid) {
+			//remember this other assignment which uses this same questionsetid but don't add duplicate entries
+			if (!is_array($qsetidassessment[$row['qsetid']]) || !in_array($row['aid'],$qsetidassessment[$row['qsetid']])) {
+				$qsetidassessment[$row['qsetid']][] = $row['aid'];
+			}
+		}
+	}
+
+	$query = "SELECT iq.id,iqs.id AS qsetid,iq.category,iqs.description FROM imas_questions AS iq,imas_questionset as iqs";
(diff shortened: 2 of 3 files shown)


drlippman:
Interesting idea.
It seems like there might be a way to avoid the repeated array_merge.  For a quick not-thought-out idea:  What about if on the first loop you just record the $qsetidassessment.  After the loop, remove any duplicate entries.  Then during the loop where $descriptions is defined, record a new array $qsetids[imas_questions.id]=imas_questionset.id. Then when iterating over the question ids, you can loop over $qsetidassessment[$qsetids[$qid]].
One other thought:  It might be better to store some sort of identifier, like "AID-1234" rather than the actual link itself, then look up the assessment name and create the link when displaying the categorization.  That way if the assessment name changes, you're not left with inaccurate categorizations.
The big potential issue here would come on course copy.  Question categorizations are copied with an assessment, and if all the categorizations are pointing to assessments in course A, then once copied to course B they'd still be pointing to assessments in course A.  So /includes/copyiteminc.php would need to be revised to modify any categorizations pointing to assessments.  One approach would be to set up something like $reqscoretrack to track every question that contains a link to an assessment, then in the doaftercopy() function, use $assessnewid to rewrite each of the question categorizations with the new assessmentid and course id.
On another important note, this line is open to SQL injection:`$query .= "AND imas_assessments.courseid=$cid ";`
Because $cid isn't processed, you're passing whatever the user provides into the query directly.  The code has some safeguards that make it safe if you put the variable in quotes:`$query .= "AND imas_assessments.courseid='$cid' ";`


 

    On Tuesday, March 8, 2016 8:41 PM, Roy Shahbazian <notifications@github.com> wrote:
 

 e.g. a review (or diagnostic) assessment can guide students to a video lesson assessment based on the student's score on a question that's in both the current assessment and another assessment. In that case, a link can be included in the Categorized Score Breakdown.Happy to make changes if there's a better way to implement it or any unforeseen issues.
You can view, comment on, or merge this pull request online at:
  https://github.com/drlippman/IMathAS/pull/23
Commit Summary
   
   - Allow categorizing questions by other assessments in which the same question appears

File Changes
   
   -  M course/categorize.php (48) 

Patch Links:
   
   - https://github.com/drlippman/IMathAS/pull/23.patch
   - https://github.com/drlippman/IMathAS/pull/23.diff

githubapprentice:
Thanks.  I'll start looking into  /includes/copyiteminc.php

course/categorize.php
```diff
@@ -96,16 +96,42 @@ function flattenarr($ar) {
...
+	$query .= "ORDER BY aid";
+	$result = mysql_query($query) or die("Query failed : " . mysql_error());
+	$assessmentnames = array();
+	$qsetidassessment = array();
+	while ($row = mysql_fetch_array($result,MYSQL_ASSOC)) {
+		//store the relevent assessment names
+		$assessmentnames[$row['aid']] = $row['name'];
+		if ($row['aid']!=$aid) {
+			//remember this other assignment which uses this same questionsetid but don't add duplicate entries
+			if (!is_array($qsetidassessment[$row['qsetid']]) || !in_array($row['aid'],$qsetidassessment[$row['qsetid']])) {
```
githubapprentice: Okay to use in_array() in this loop?

assessment/catscores.php
```diff
@@ -86,6 +86,13 @@ function printoutcomes($arr,$ind,&$outcomenames, &$catscore, &$catposs) {
 	foreach (array_keys($catscore) as $category) {
 		if (is_numeric($category)) {
 			continue;
+		} elseif (0==strncmp($category,"AID-",4)) { //category is another assessment
+			$categoryaid=substr($category,4);
+			$query = "SELECT name FROM imas_assessments WHERE id='$categoryaid' LIMIT 1";
```
githubapprentice: Is it preferable to add "AND courseid='$cid' " to the query?

githubapprentice:
Did I mess up outcomes after a course copy?  After copying a course with no custom outcomes but categories set to assessments, there are a couple warnings about outcome arrays being empty (e.g. copyitems.php near line 220, addoutcome.php line 144, categories.php line 78).  Not sure how to prevent it.

includes/copyiteminc.php
```diff
@@ -461,6 +477,20 @@ function doaftercopy($sourcecid) {
...
 	}
+	//update any assessment ids in categories
+	if (count($categoryassessmenttrack)>0) {
+		foreach ($categoryassessmenttrack as $newqid=>$oldcategoryaid) {
+			//is oldcategoryaid in copied list?
+			if (isset($assessnewid[$oldcategoryaid])) {
+				$query = "UPDATE imas_questions SET category='AID-{$assessnewid[$oldcategoryaid]}' WHERE id='$newqid'";	
+				mysql_query($query) or die("Query failed : $query" . mysql_error());
+			} else if (!$samecourse) { //since that assessment isn't being copied, unclear what category should be
+				$query = "UPDATE imas_assessments SET category=0 WHERE id='$newqid'";
```
githubapprentice: If the assessment wasn't copied, I assumed category should be 0.  Let me know if that's not the best choice.

githubapprentice:
On further testing, it looks like the warnings from the outcome arrays being empty are not related to this pull request.  Copying a course with no outcomes and no assignments as categories also gives the same warnings. I'd be happy to a add a check for a non-empty array before the foreach statements.

assessment/catscores.php
drlippman: It would probably be a good idea, just to catch any cases where the $categoryaid is, by some mistake, pointing to an assessment outside of the course.

course/categorize.php
drlippman: Yeah, it's fine.  Since you don't need the imas_questions.id, you _might_ be able to eliminate the need for it by changing the SELECT to:

`SELECT  DISTINCT imas_questions.questionsetid AS qsetid,imas_assessments.id AS aid, imas_assessments.name`

I think that would give you distinct qsetid/aid sets, but you'd want to test it out.

course/categorize.php
```diff
@@ -96,16 +96,42 @@ function flattenarr($ar) {
 		$questionlibs[$row[0]][] = $row[1];
 		$libnames[$row[1]] = $row[2];
 	}
-	$query = "SELECT iq.id,iq.category,iqs.description FROM imas_questions AS iq,imas_questionset as iqs";
+
+	//add assessment names as options
+	//find the names of assessments these questionsetids appear in
+	$query = "SELECT imas_questions.id AS qid,imas_questions.questionsetid AS qsetid,imas_assessments.id AS aid,imas_assessments.name ";
+	$query .= "FROM imas_questions, imas_assessments ";
+	$query .= "WHERE imas_questions.assessmentid=imas_assessments.id ";
```
drlippman: It's equivalent, but I think it's considered slightly more correct to use an explicit JOIN:

````
SELECT (blah) FROM imas_questions INNER JOIN imas_assessments 
ON imas_questions.assessmentid=imas_assessments.id 
WHERE imas_questions.questionsetid = ANY (SELECT ...
````

File: includes/copyiteminc.php (hunk 3 above)
Comment (drlippman) [inline includes/copyiteminc.php]: Yeah, 0 still is the default "no category" value, from back when category was a numeric field.

**Comment by drlippman** [comment]
Yeah, I have a nasty habit of forgetting to add checks for empty arrays.  If you want to add them, that'd be lovely, or I can try to get around to it :)

This is looking really good.

**Comment by githubapprentice** [comment]
Good call on the DISTINCT rather than in_array().  It tested fine.

I'll do another pull request for the empty array checking.

File: assessment/catscores.php [hunk 5]
```diff
@@ -88,7 +88,7 @@ function printoutcomes($arr,$ind,&$outcomenames, &$catscore, &$catposs) {
 			continue;
 		} elseif (0==strncmp($category,"AID-",4)) { //category is another assessment
 			$categoryaid=substr($category,4);
```
drlippman: Will you please make one more small change.  Change this line to
`$categoryaid= intval(substr($category,4));`

I think someone could try to inject SQL here by creating a custom category like `AID-4'; DROP TABLE imas_courses; --`.  Rather than escape $categoryaid, since we know the categoryaid is supposed to be an integer we can use intval to prevent this.

githubapprentice:
Yes, my bad for not catching things like that.

Also, if you think it's reasonably likely someone would have their own outcome named "AID-...", I'd be happy to change this if() condition to be stricter, e.g. /^AID-[0-9]+$/

drlippman:
I don't think that is necessary.
On Mar 17, 2016 11:40 AM, Roy Shahbazian <notifications@github.com> wrote:Yes, my bad for not catching things like that.

Also, if you think it's reasonably likely someone would have their own outcome named "AID-...", I'd be happy to change this if() condition to be stricter, e.g. /^AID-[0-9]+$/

drlippman:
I was describing this to a colleague, and they had a good idea:  On the Add/Remove questions page, when you're using the "Choose from Assessment" option, there should be a way to auto-setup the assignment categorization when you add questions.  Perhaps a checkbox on the addset question options page that shows when you click the Add button?  

I'll go ahead and merge in this request, though, and if you want to add that later, you can.  Or I can :)

PR #24, opened 2016-03-17
Prevent iterating over empty outcome arrays

githubapprentice:
I used this syntax to avoid indenting the entire block so it has a clean line-diff.  Happy to use another syntax if preferable.


drlippman:
It looks like in all three cases the issue is `$outcomes = unserialize($row[0])` returning a non-array.  I think the cleaner fix might be just add a check after the unserialize, and initialize $outcomes as an array if it's not already one.  I'll go ahead and make that change.

PR #25, opened 2016-03-18
Added quick pick first assessment category

githubapprentice:
This isn't yet quite as automatic as you suggested, but a start.  I'll still look into having this kind of functionality on the Add/Remove Questions page.

Let me know if the user interface would be better simply having two separate buttons rather than a select box.


drlippman:
Looks good.

PR #28, opened 2016-05-24
took away strict mode for mysql sessions, and added a block header id…

cooplogic:
… for all header tags. The first change just adds compatibility to newer versions of mysql, and the second change allows a teacher to more easily make internal links between blocks by adding id anchors.

### DIFF
--- a/config.php.dist
+++ b/config.php.dist
@@ -197,6 +197,9 @@
 	  unset($dbserver);
 	  unset($dbusername);
 	  unset($dbpassword);
+	  $query = "set session sql_mode=''";
+          $result = mysql_query($query);
+
   }
   //clean up post and get if magic quotes aren't on
   function addslashes_deep($value) {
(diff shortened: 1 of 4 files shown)


drlippman:
These look fine.  But it's got me wondering - I wonder if the "better" solution is to allow NULL on the text fields.  I believe that would also potentially save some memory.  Is there any reason having null fields would an issue?

cooplogic:
I think allowing NULL would probably be better, but you need to add error checking for NULL, if you allow it. There is also a numeric field that isn't set on one of the inserts or updates. I think you would want to do more testing if you start to allow NULL fields.

David
Cell/Text: 215-516-9858

Sent from my iPhone

On May 23, 2016, at 11:41 PM, David Lippman <notifications@github.com> wrote:

These look fine. But it's got me wondering - I wonder if the "better" solution is to allow NULL on the text fields. I believe that would also potentially save some memory. Is there any reason having null fields would an issue?

PR #30, opened 2016-06-07
 added automatic preview for caculate( and AMpreview( functions

cooplogic:
This code changes inputs with an AMpreview or a caculate preview to update in real time as you type. MathQuill is turned of in this case, since the popup for the mathquill interferes with the preview.


drlippman:
I've considered adding this on several occasions, but have always resisted, since it results in a lot of "syntax error" messages displaying during the typing/entry process, which I fear will confuse/worry students more than the auto-preview will help.  I also really don't like the idea that this disables the MathQuill option.

I'd be willing to consider adding something like this, but first off it would need to not disable the MathQuill input.  Second, it might be better to put it on a timer, so that it doesn't trigger the preview until a few seconds after the student stops typing.  Or maybe do a real-time preview without syntax check, then trigger the syntax check only a few seconds after the students stops typing.  

Maybe something like this as an oninput function:
````
var livepreviewtimer = null;
function livepreview(inputId,outputId) {
	var outnode = document.getElementById(outputId);
	var n = outnode.childNodes.length;
	for (var i=0; i<n; i++)
		outnode.removeChild(outnode.firstChild);
	
	var dispstr = document.getElementById(inputId).value;
	outnode.appendChild(document.createTextNode('`'+dispstr+'`'));
	if (!noMathRender) {
		rendermathnode(outnode);
	}
	clearTimeout(livepreviewtimer);
	livepreviewtimer = setTimeout(function() { AMpreview(inputId,outputId);}, 1000);
}
````

The re-render flicker is pretty annoying, though.  Maybe use the buffering technique from [this site](https://cdn.mathjax.org/mathjax/latest/test/sample-dynamic-2.html) to reduce that.  And maybe set the outnode.style.position="absolute" during preview to prevent the mathquill and tooltip boxes from drifting.

cooplogic:
I added the changes that you suggested. I modified AMhelpers_min.js since that is the one that is actually used. I wasn't sure how you get the minimized version from AMhelpers.js.

In order to get mathquill to work and to make it so that the hints and other buttons wouldn't interfere, I added some whitespace. I think there is a more elegant way to do this, such as put the span after the Hint button and other help button creation, but I couldn't find where that happens.

The coursereport page is for a different purpose entirely, and can be ignored for now.

cooplogic:
The livepreview will now update when mathquill saves. Please let me know if there is anything I can do to help getting this pulled upstream.

drlippman:
There are still a number of issues to be considered and addressed, particularly around how this interacts with other display approaches (images, katex, etc), and trying to integrate the buffering approach to reduce rerender jitter.  I am personally more interested in getting an always-on MathQuill entry working, but will try to find some time to dig into this and see if I can polish up some of those details.

cooplogic:
Do you have an example course, or example problems that I could load into my local version of IMathAS so that I could see how the interaction fails with other display approaches. With the absolute positioning, there doesn't appear to be much, if any, rerender jitter, but I can look into buffering as well, if you think it's needed. 

For the always on MathQuill, are you suggesting an embedded mathquill window?

drlippman:
My main concern with the image-based display was how quickly the image server would get bombarded with render requests :) Perhaps a longer delay when in image-based mode would do it.  

I like that in the buffer demo you don't get that MathJax "twitch" where the original input code and/or preview render display before the actual render. If it's not too horrible to integrate that, it seems like it'd be nice.

For the update on MathQuill insert, I noticed you added another input element. Would it work to just call `$("#"+cureedd.id).trigger("input")` on the insert instead of addition another DOM element? I guess I can test it myself later :)

The always-on-MathQuill would be a MathQuill input field in place of the regular text entry field. Implementing that is pretty trivial - the hard part has been figuring out how to provide a usable symbol input palette in an unobtrusive way, particularly on mobile. For mobile I'm trying to emulate the idea of the input panel on desmos and geogebra. For desktop I'll probably have a palette docked on the right side of the screen. But all that work has been very slow going, as I keep getting distracted with other things, like my classes ;)

cooplogic:
It looks to me like the twitch comes from this code: 

`    outnode.appendChild(document.createTextNode('`'+dispstr+'`'));
    if (!noMathRender) {
        rendermathnode(outnode);
    }
`
I think that the problem is that rendermathnode uses the textnode with the string that needs to be converted. Perhaps if rendermathcode used a different kind of node that was hidden by default as the initial string, and then populated the node with different visible children, then the twitch would go away.

I am unsure of how to convert the MathJax buffering code to work with the livepreview code, since in their example the Preview class appears to be called directly without parameters. It's not exactly the parameters that are the problem, but the fact that it looks like we would need multiple objects with the Preview class if there were more than one box that needs to be previewed, and I'm not sure of the best way to make that work in javascript.

drlippman:
> Perhaps if rendermathcode used a different kind of node that was hidden by default as the initial string, and then populated the node with different visible children, then the twitch would go away.

Yup, that's the whole idea of that buffering solution I mentioned :)  It's not necessary for typical renders, but would be necessary for something like this live preview.

From what I can see, an analog to rendermathnode would need to be developed for live preview purpose, with different functions for each display method (MathJax, Katex, images), since the mechanism for doing it "right" will change with each. 

I'm happy to work on that as my time allows. But I have several items of higher priority on my plate, so it might be a while.

drlippman:
I managed to get an initial version of the buffered live preview working for all render modes. I still have to enable it on all the question types, but if you want to take a look, it's in the [feature.livepreview](https://github.com/drlippman/IMathAS/tree/feature.livepreview) branch

cooplogic:
Thanks. I will take a look next week.

cooplogic:
This looks good! The preview button is still there. If you'd like me to help enable it for all questions types, I could try my hand at it, if you can point me to the code that enables one question type and the code that still needs changes, I'd be happy to work on it.

drlippman:
Rewrite merged in a70a098f5ba93f61072b7e35f747771a6b05e391

PR #31, opened 2016-06-13
Instructor view can see and copy link to each block.

cooplogic:
This change is to make it easier to make a local link to a block in a course.

### DIFF
--- a/course/courseshowitems.php
+++ b/course/courseshowitems.php
@@ -163,7 +163,9 @@ function enditem($canedit) {
 						echo "<a href=\"addblock.php?cid=$cid&id=$parent-$bnum\" $astyle>", _('Modify'), "</a> | <a href=\"deleteblock.php?cid=$cid&id=$parent-$bnum&remove=ask\" $astyle>", _('Delete'), "</a>";
 						echo " | <a href=\"copyoneitem.php?cid=$cid&copyid=$parent-$bnum\" $astyle>", _('Copy'), "</a>";
 						echo " | <a href=\"course.php?cid=$cid&togglenewflag=$parent-$bnum\" $astyle>", _('NewFlag'), "</a>";
-						echo "\n<br> Link to block: &lt;a href=\"#blockh{$items[$i]['id']}\"&gt;{$items[$i]['name']}&lt;/a&gt;";						
+						if ($items[$i]['avail']==2 && strpos($parent,"-") === false ) {
+						  echo "\n<br> Internal link to block: &lt;a href=\"#blockh{$items[$i]['id']}\"&gt;{$items[$i]['name']}&lt;/a&gt;";
+						}
 						echo '</span>';
 					}
 					if (($hideicons&16)==0) {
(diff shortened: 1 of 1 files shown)


drlippman:
There are several potential problems I see with this.  First, this creates a link within a page, so the link is useless from anywhere except the course page (messages, forums, etc.).  Second, the link won't work if the block is hidden or hasn't been dynamically loaded (because it's nested inside a closed block or folder).  Third, block IDs are not preserved on course copy, so when the course is copied all the links would not work properly anymore.

These sorts of challenges are the big reason I haven't added any sort of internal course item linking ability into the system :)

cooplogic:
I see your point that simply adding a way to link to each block is not a complete solution to making a course internally linkable. However, it is a start, and this incremental change can lead to better navigation within the course. 

To address your first point, the purpose of this link is for internal navigation, and is not meant to address linking from other pages. Perhaps the text should read "Internal Link for block: " instead of "Link for block:"

Second, I hadn't considered nested or hidden blocks, but I could modify the code so that the links only display for "showing always" blocks that aren't nested. Perhaps a better long term solution would be to create a symbol that is used to specify links, and then the symbol can be parsed by the php and replace the link dynamically based on that symbol. Then if the link wasn't reachable, it wouldn't be created.
 
Third, the fact that the link will break because the block IDs aren't preserved on course copy definitely creates more risk. Anyone copying courses would need to be made aware of this risk if there are internal links. However, much of this risk can be minimized by creating the links only immediately after the course is copied. That way the block ids will be preserved on copy. Later, this could be fixed by preserving the block IDs programmatically. 

In the short term, the links could be used with a caveat to check and verify that internal links work after copying the course. In the long term, preserving block ids in each course programmatically would solve much of the course copying problems. However, a user would still need to check any links that were copied from a different course to make sure that the internal link goes to the appropriate place.

cooplogic:
The latest commit addresses your second concern.

drlippman:
First off, let me say thank you for submitting this and your other pull requests - I really do appreciate getting suggestions for improvements in code form :)

That said, since most of the users of the system are fairly non-technical, I really try to avoid features that only work sometimes, or with caveats. I'm imagining, for example, a course with one of these links, and a student clicking it and nothing happening; that would be very frustrating and confusing for them.  I'm going to have to hold off on this idea until it can be implemented in a way in that can work consistently and reliably from any context.  

I've had a few different ideas about how to handle courselinks, and don't love any of them.  One idea is to have any course links point to a special linkresolver.php page, which would resolve the course link and redirect the user to the correct page.  That could take the block ID, resolve it into a folder string it could feed to course.php to take the user directly to that block's contents.  Optionally, a javascript function could scan for folder links, and rewrite them "#block1234" style hrefs if the folder was already displayed on the page.  

Of course, any of these links would need to be tracked and updated during a course copy, including removing any links that point to items that aren't copied.  It's the pile of additional details that have prevented me from wanting to tackle this issue :)

On a side note, all the rest of that was gotten to work right, it would be neat to add an editor plugin that would provide a pop-up tree of the course content, so a user could insert a course link by selecting from the list, rather than having to hunt down the link somewhere.

cooplogic:
Ok, I'll work on a more holistic version of the course linking. The majority of the changes I've added are based on the needs of the Bryn Mawr College team that are making a set of courses for a study using MyOpenMath. The goal is to have the features as soon as possible, while also minimizing changes that could adversely affect other users of the system.

drlippman:
If you need something immediate, and you think that your version of linking is sufficient, you might just implement it as a "bookmarklet" for your team to use.  For example, if your team put this as the URL for a bookmark in their browser, clicking it would expose the same "Link to block" link you were wanting.

````
javascript:$("span[onclick^=toggleblock]").each(function(i,el) {var blockid = $(el).find("a").attr("id"); var blockname = $(el).find("a").text(); $(el).siblings(".instronly").append(' | Link to block: <a href="#'+blockid+'">'+blockname+"</a>");});
````

cooplogic:
That works great with a simple edit: I changed < to &lt; and > to &gt; so that the link didn't autoexpand on copy. Thanks! I will continue thinking about a more robust solution.

cooplogic:
I'll make this request later if I come up with a more robust solution.

PR #32, opened 2016-06-14
added collapse all button in left sidebar

cooplogic:
This creates a collapse all button on the left sidebar to clean up a course after expanding many blocks.

### DIFF
--- a/course/course.php
+++ b/course/course.php
@@ -19,7 +19,7 @@ if (!isset($teacherid) && !isset($tutorid) && !isset($studentid) && !isset($gues
 	$body = _("You are not enrolled in this course.  Please return to the <a href=\"../index.php\">Home Page</a> and enroll\n");
 } else { // PERMISSIONS ARE OK, PROCEED WITH PROCESSING
 	$cid = $_GET['cid'];
-	
+	$collapseall = $_POST['collapseall'];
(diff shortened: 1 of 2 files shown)


drlippman:
If your course is getting messy because of too many expanding blocks, then students will probably find that challenging too. It might be best to approach this from an instructional design perspective, and redo your course using more folder-style blocks rather than just expand/collapse blocks.

That said, as for the idea itself I can definitely see the value. However I've been trying really hard to clean up the UI, so adding something to the left menu requires something super important and essential, which this doesn't qualify as for me.  I wonder if it could be implemented in some way that doesn't require an additional UI element.  Perhaps shift-clicking the collapse toggle for one block could collapse them all?

Finally, collapsing the blocks shouldn't require a POST action; it should be doable with javascript alone.  

````
$("a[id^=blockh]").each(function(i,el) { 
   var id=$(el).attr("id").substr(6); 
   if ($("#block"+id).hasClass("blockitems")) { toggleblock(id,null);}
});
````

Looking at this did make me realize I really could use some better use of semantic classes on DOM elements (my excuse is that most of that code was written 10 years ago when I was still learning).

drlippman:
Note:  The change I just pushed will prevent the javascript snippet above from working (because the arguments to toggleblock change), in case you were using it as a bookmarklet.

PR #33, opened 2016-06-14
Weekly Report

cooplogic:
Added a weekly report so that teachers could get a quick summary of recent activity in their courses. The tables still need some css for formatting. Any feedback you have would be appreciated.


drlippman:
I haven't fully gone through this yet, but from an initial look it appears you're trying to report on a lot of things the system doesn't actually track enough detail to report. I think you'd get a lot of misleading information from some of these reports. 
On Jun 14, 2016 12:09 PM, David Cooper <notifications@github.com> wrote:Added a weekly report so that teachers could get a quick summary of recent activity in their courses. The tables still need some css for formatting. Any feedback you have would be appreciated.



You can view, comment on, or merge this pull request online at:
  https://github.com/drlippman/IMathAS/pull/33

Commit Summary
simple weekly reportadded code to show all students in student summarystudent summary tablesadded assessment tablesmodified some text for clarity

File Changes

    A
    course/coursereport.php
    (557)
  
    M
    index.php
    (4)
  

Patch Links:
https://github.com/drlippman/IMathAS/pull/33.patchhttps://github.com/drlippman/IMathAS/pull/33.diff

drlippman:
Also, it looks like you're building this on top of the main course display code. It really should be a distinct standalone page.
On Jun 14, 2016 12:09 PM, David Cooper <notifications@github.com> wrote:Added a weekly report so that teachers could get a quick summary of recent activity in their courses. The tables still need some css for formatting. Any feedback you have would be appreciated.



You can view, comment on, or merge this pull request online at:
  https://github.com/drlippman/IMathAS/pull/33

Commit Summary
simple weekly reportadded code to show all students in student summarystudent summary tablesadded assessment tablesmodified some text for clarity

File Changes

    A
    course/coursereport.php
    (557)
  
    M
    index.php
    (4)
  

Patch Links:
https://github.com/drlippman/IMathAS/pull/33.patchhttps://github.com/drlippman/IMathAS/pull/33.diff

cooplogic:
It is a standalone page, but I did use the header code from the course.php page to keep the look and feel.

The reports just get activity in the last week. Here is an example:

image

drlippman:
There's still a lot of code in there that is unnecessary for the purposes of the page.

There is no way in the system to identify that an attempt was made in the last week.  It's only possible to determine that the assessment was last modified in the past week. 

Looking at your sample report, I'm really unclear on what it is supposed to be conveying.  Can you give me the big picture of what you're trying to capture with this?

cooplogic:
I did my best to extract only the necessary code for the header, but if it can be cleaner, then I'll be happy to take it out.

From what I understand, the imas_assessment_sessions reports its time each time a user accesses an assessment. 

The reports should give an idea of the activity of the students each week. It will say which assessments accessed in the last week have been given credit, and which ones have not been given credit. It will summarize with respect to the users and the assessments that they've accessed as well as with respect to the assessments and the users that have accessed the assessments in the last week.

drlippman:
Each individual action's time is not logged in imas_assessment_sessions.  Only the initial start time and last submission time are recorded, so it's hard to know what level of activity was done inside the assessment during that action.

I worry that this report is too specifically tailored to how you are structuring your courses and the evaluation in those courses.  For example, most people don't use the "minimum score to receive credit" setting, so this report would make no sense to them.

drlippman:
I've been giving this some more thought. Perhaps it would be reasonable to have a place in the system where there are a collection of reports available, and the user could select one appropriate to their course. This one you've been working on could be a "Lab Course Report" or something like that, appropriate for courses using "min score". There could eventually be another "At Risk Report", something I've considered for a while, showing students who have habitually started homework assignments at the last minute. I'd probably put that link somewhere from the Gradebook though, maybe possibly from the left navigation, but definitely not from the front course list page.

Some additional comments/ideas:

- You don't need to include courseshowitems, htmlutil (unless you're using it), and calendardisp.  

- Everything from line 49 to 108, and 130-194 could get replaced with a generic breadcrumb load; the main course.php page does a bunch of stuff that normal pages don't need.  Likewise, everything from line 409 to the end is not needed for your page.

- Is there a reason to use group_concat then explode the results, rather than just pulling each record individually and adding them to your array?

- I'd recommend adding a date-picker or pulldown or something to allow the user to select a date range, rather than forcing a week.

- Rather than requiring SQL to do a date conversion on every date, it'd probably be easier to precompute `$rangestart = time() - 7*24*60*70` then have your WHERE be `endtime > $rangestart`.  This would give the flexibility to adjust that time interval for the range more easily.  Also, I don't think it makes sense to use `greatest(starttime,endtime)`, since endtime would only be smaller if the student hadn't submitted any work yet.  In that case, I doubt it really  makes sense to count that as an "attempt" at the assignment.

- I'm imagining with a bunch of students and a bunch of assignments, this report could get really messy really quickly.  It might make sense to break it into two separate reports - a by-student report and a by-assessment report.

- You might be able to condense/cleanup the display a bit by combining the student summary and student detail sections, by making the list of assessments a student got credit for show as a pop-up or hover-over on the number list.

cooplogic:
I made some changes based on your feedback, and I had forgotten to compute the possible scores, so I fixed that as well.  courseshowitems.php has the function getpts, which I use.

I used group_concat initially because my goal was to create each table from a single database call. The problem is that both assessment scores and max scores are not stored in the database, but they have to be computed each time.

A date picker might make sense, but once you have chosen an interval, it probably makes sense to always use the same interval. That way if you want to save the table data into a spreadsheet to keep track of weekly (or pick your range) progress, then you just need to run the report at that point.

I use `greatest(starttime,endtime)` because (empirically from testing) it looks like starttime is used when the first attempt is made, but endtime is used afterwards.

I like the idea of precomputing the range.

The reason to put this on the main page is to use this report as a dashboard feature of the course. That way you just have to log in, and you can click once to get the report. 

I'll need to think some more about the best way to make efficient database calls.

As far as the results looking messy, the goal is that the numeric data will be used first to identify the student or the assessment of interest. Having a link from one to the other could make good sense.

Here is a screenshot of the latest commit:
image

drlippman:
I have a feeling using a fixed 50% breakpoint is not going to be generally useful either.  For assessments with minscore for credit set, it makes sense to use those.  For assignments without it, perhaps a user-selectable breakpoint would make sense.

I could see this evolving into a collection of reports, where the instructor can choose their default report and report settings for the course based on the course structure.

I'm not going to add links to this to the main course list, at least not the way the current course list works. Eventually I'd like to add a drop-down menu for each course for accessing settings and such (the things currently on the Admin page), and if I ever get to that, the reports page could get listed in that drop-down. Until then, reports would make more sense as a sublink off the Gradebook, or possibly as a left-navigation item under Tools. 

> starttime is used when the first attempt is made, but endtime is used afterwards

starttime is recorded when the assignment is opened.  endtime is recorded when the assignment is first submitted.  So endtime is a better measure of an "attempt" at the assignment.  starttime just indicates they clicked the link to open it.

> but once you have chosen an interval, it probably makes sense to always use the same interval

So maybe have a pull-down for interval options instead of a date-picker.  I'm just thinking some people might want a "this week" view, showing activity since Monday, in which case a fixed interval is less useful.

> courseshowitems.php has the function getpts, which I use.

That's a lot of code to load for one function.  One of these days I'll break out the utility functions into a separate file, but for now just copy that function into your code.  Your code reminds me I really need to denormalize the database a bit and store the total points possible for an assignment directly in imas_assessments.

On an unrelated note, you might try popping a class="gb" on your tables to give them a little system-default styling.

cooplogic:
I think that a link from the left panel would be better than a link from gradebook.

I've made changes based on your suggestions. 

I feel like this is a good first step to the reports. Is there anything else needed before you can accept this pull request?

cooplogic:
Here's what the latest version looks like:

image

drlippman:
As I mentioned before, just linking to this page from the main course page isn't going to work for me, since this page is not relevant to most users.  The bare minimum for me to consider merging this would be taking out the index page link, and putting in a course page nav link to an intermediary "reports" page, which then has a link to this report.  If you want to get fancy, you could add a "make this my default report" option so that the report link automatically redirects to this report, but that's not necessary

Also, you should take out the maketopmenu code which isn't needed.

I think there's a lot of room for improving the visual look of the page.  The [name]:[name] way of displaying assessments is pretty klunky.  Maybe use a proper `<ul>` or something to make it more semantically meaningful and easier to read?  

As I mentioned before, the two first student summary view seems kind of redundant with the second. A simple page-condensing solution might be to put the <ul> of assessments each student attempted hidden under the number count in the top report.  Then clicking the number would expose the <ul> of assessments, and you could add a "show assessments attempted" button to expose all of them.

One more thing: Displaying usernames is not going to be very informative to most teachers. It'd be better to grab their actual name from imas_users.

drlippman:
Just as a note:  Please understand that the master branch of the code is the code that is live on myopenmath.com and wamap.org, so "first steps" need to be sufficiently flushed out to go onto production sites before I'll consider merging them.

drlippman:
Is there a reason you're including imas_students in your student summary query?  It seems unnecessary.  

It strikes me that it should be possible to generate both summary views from one non-grouped database pull, perhaps by just modifying your second pull to include ias.bestscores.  It might be necessary to do a separate pull to get the imas_assessments info, but that might be better than doing the two pulls again imas_assessment_sessions.

drlippman:
Note to self: If this gets merged, an index on imas_assessment_sessions.endtime needs to be added.  The main query crashed my dev server.

cooplogic:
The imas_students join is there because the following joins are left joins, and only students in the course should be included, not all users.

Aside from the click to show behavior table combination, I think I've addressed all of your concerns. I'll look at the click to show on Monday.

Thanks very much for your feedback. I do understand that the code needs to be sufficiently fleshed out before being incorporated.

drlippman:
Ahh, right, since you're not pulling the student list separately, you need imas_students to filter out the teacher attempts.  I wonder whether doing the extra join and pulling the student's name on every single row is worth avoiding a second database call?  (I really don't know)  

I am concerned by that "or" in the where clause; it seems like that would mean we'd be pulling records that were not from the intended course.  I tried running that query on my dev server and it crashed it.  I think it might be cleaner and safer to just pull the student list in a separate database call.  It's possible adding a `and stu.courseid='$cid'` to the join would be enough to solve it, but I really do think separating it into a different call is probably cleaner.

While what you did is OK, in the future you might consider using mysql_fetch_assoc so that you don't have to use variables to store numerical indexes.  I've learned over the years that using numerical results makes maintainability much more challenging on queries with more than 2-3 columns.

I haven't gone through the code in detail yet, but from a first glance it looks good.

cooplogic:
I cleaned up the $st table creation so that it happens in 2 steps. First the imas_students and imas_users table is joined, and the first and last name are obtained for each student in the course. Then the imas_assessments and imas_assessment_sessions are joined to get the bulk of the data needed for the table.

In addition, now there is only one table for the student summary and one table for the assessment summary with the ability to click a number to get more detail on the no credit and credit columns.

Thanks for the feedback. This is definitely much cleaner than before.

drlippman:
Thanks.  I'll probably merge this into a branch and do some additional cleanup/tweaking before merging it into master.

drlippman:
This has been merged into the [feature.reports](https://github.com/drlippman/IMathAS/tree/feature.reports) branch.  As I went through your code, I had difficulty deciphering the intent of some of your variable names (as I'm sure people do with my code), so I ended up refactoring it quite a bit.  Let me know if I've mangled anything in the process.

I'd really like to get another report type added, and possibly expand this report with some customization options, but if you're in a hurry and need this report available this summer, let me know and maybe we can get it up earlier.

cooplogic:
It looks good. I like the row based expand collapse and the expand all collapse all buttons. I'm not attached to variable names. 

There are a few things that aren't working quite right:

1. The credit/no credit doesn't appear to be working correctly. I think it's because you have line['minscores'] instead of $line['minscore'].  (line 164)
2. Cumulative score uses totalPointsOnAttempted (the numerator) to decide whether or not there is a cumulative score, when it should be totalPointsPossibleOnAttempted. (line 272)
3. The rangestart is not being set to strtotime("1 week ago") instead it is being set to 0, so all assessments are showing up. I saw a note about time zone handling. Assuming the web server is in the same location as the database this call should work correctly regardless of the client's time zone. If not, then it would be less error prone to use the database time functions.

I would like this report to be available this summer, since the teachers who will be using them might have feedback for how to make them more useful, and I'd like to get those changes in before the fall.

drlippman:
Thanks. I'll get those fixed.

For the timezone issue, some of it depends on what you want "1 week" to mean.  Literally 7*24 hours from when the report is generated?  Or from the start of the day a week ago?  

The timezone issue is more relevant to the "since monday" type of intervals, since then the start timestamp would depend upon the user's timezone, not the server's.

PR #35, opened 2016-06-23
Add Solver tool to aid in question-writing

githubapprentice:
Thanks for the suggestion on redefining console.log to catch the completion of sagecell evaluation.  Worked great!

The popup is movable but there's a few pixels of offset when you first drag.
A few other possible TODOs: evaluate on paste, move JavaScript and Help to other files.
I haven't tested it on Chrome or IE.  Drag-drop expressions doesn't work on Android, but otherwise okay.
Sorry for all the extra merges commit-reverts here.  I probably did something wrong with branching.  Let me know if you know how to clean up the log.
I'm happy to work on any other suggestions.


drlippman:
No worries on the commit log.  Thanks for this. I look forward to digging into it soon. I do think it'd probably make sense to split all the JavaScript into an external file so it can be cached.

githubapprentice:
This moves the solver JavaScript, css and help text out of moddatatext.php.  Let me know if these file locations should be different.

drlippman:
Super, thanks.  This is so darn slick.  Incredibly well done.

A few ideas/thoughts as I play with it:
- When your system auto-detects the operation to perform, it'd be nice to select it in the solveroperation pulldown so it's clear what was auto-detected
- For cases where the user needs to hand-edit the Sage, it'd be nice to have a "re-evaluate" button.  Maybe it could be made to appear using an onchange callback `CodemirrorInstance.on('change',function(cMirror){ make button appear});`
- If the output produces an error (I think `.match(/Traceback/)` would do it), then don't put the output in the "drag or copy this" area.  Maybe report an error occurred instead?
- If an expression contains the same variable multiple times, we don't need to define the variable multiple times, like `x,b,x,c,a=var("x,b,x,c,a")`.  
- For the output, you say "drag or copy", but there's no way to select it to copy it, so it might be better to say "drag or use the buttons below"
- The output doesn't seem to reinsert the $ on variables.  I think you have an extra backslash in your `\\\b` regex.
- For the dragging jump, change line 72 in the mouseup to `$("#solverpopup").position();` rather than `.offset()` to fix that.
- Is there a reason for the light blue color for the title bar?  Seems a bit light compared to the white text.  I'd also suggest maybe not overriding default/theme fonts unless there's a strong reason to.

drlippman:
Oh, and I noticed that on a fixed-width theme, the solver is limited where it can be dragged.  It would probably be good to move the solverpopup HTML into a `$placeinfooter` variable, which will get inserted by footer.php outside the limiting .midwrapper element.

githubapprentice:
Thanks for testing and suggestions.  This is an attempt to address those issues.

When the user drags text from the question code to the solver, should it delete what was already there?
If you know how to catch the paste event for the solver input box, I'd like to have it trigger evaluate.
Good catch on the \\\b.  My approach with escaping was to just keep adding backslashes until it worked.  And if two backslashes is good, then three backslashes should be even better, right? :)
I don't think I have a fixed-width theme to test that.  Let me know if it's still a problem or if the initial popup position should be different.

drlippman:
Looks great!  I might make a couple small tweaks to the colors (I think I liked the white background better) but should have it merged shortly.

drlippman:
For sometime in the future, it looks like 
````
expr = a/(x+b)+c/(x+d)
expr.simplify_full()
````
actually does a decent job of simplifying/combining stuff.  It would require a refactor of how you do your operation changes, though, since it doesn't fit the other patterns.

drlippman:
Oh, here's a way that could minimize work:
Add a line at the top, `def simpfull(expr): return expr.simplify_full().simplify_trig()`.  Then you can call that like `solve` or `diff`.

PR #39, opened 2016-07-06
change embedded frame post message to a json string

bracken:
This allows the data to be parsed a bit easier on the receiving end.

You can receive these messages with something like:
```
  window.addEventListener('message', function (e) {
    try {
      var message = JSON.parse(e.data);
      switch (message.subject) {
        case 'lti.frameResize':
          // resize frame using message.height
          break;
        case 'lti.ext.mom.updateScore':
          // handle score
          break;
      }
    } catch (err) {
      (console.error || console.log)('invalid message received from ', e.origin);
    }
  });
```

### DIFF
--- a/OEAembedq.php
+++ b/OEAembedq.php
@@ -153,7 +153,7 @@ if (isset($QS['showscored'])) {
 	
 	echo '<script type="text/javascript">
 	$(function() {
-		window.parent.postMessage("updatescore='.$signed.'","*");
+		window.parent.postMessage(JSON.stringify({subject: "lti.ext.mom.updateScore", jwt: "'.$signed.'", frame_id: ' . $qsetid . '}), "*");
 	});
 	</script>';
 	if ($scoredonsubmit) {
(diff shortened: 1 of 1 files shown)


drlippman:
Since I can't imagine the iframe in OEA would actually have an ID just consisting of the question ID, I wonder if it would make sense to make the frame_id (or element_id, to be consistent with the Candela resizer) something like `"mom-oeambedq-".$qsetid`.  That way you wouldn't have to rewrite the lti.frameResize frame_id value when receiving the message - you could just use it directly as the ID like you probably do for other embeds.

bracken:
the candela one doesn't actually use that key right now, so I renamed it to `frame_id` which I like better.
The idea would be to actually pass the id in to the embedded doc so the iframe owner can own the ID too. So what I'd do is have a `&frame_id=mythingiwant` and then the post message would resend that same thing.
I figured I could add that in later.

drlippman:
Gotcha.  I did something similar for the multiembedq page.  I'll merge this, then probably

PR #40, opened 2016-07-07
Change to the DBsetup file 

AbeerKhakwani:
# what 
I made a change to the dbsetup.php file.  #brackenapproved? 
# why 
The imas_questions query was erring on the install. 
# change 
 Made the```showans` CHAR(1) UNSIGNED NOT NULL DEFAULT \'0\','``` to  ```showans` TINYINT(1) UNSIGNED NOT NULL DEFAULT \'0\','```


bracken:
I had this problem too when trying to create a new DB.

drlippman:
Thanks. It's actually supposed to be `CHAR`, I just forgot to remove the `UNSIGNED` from the dbsetup file when I changed it.  I'll get it fixed.

PR #43, opened 2016-07-15
new testtype EndReviewWholeTest

cooplogic:
This creates a new test type where users must retake the whole test, but can review their last attempt immediately upon completing the assessment.


drlippman:
At first I was very confused by this :)  But I think I see what you're getting at:  a way for students to do an assignment, see their results at the end, then (using the reattempts different versions option), allow them to reattempt the whole test, essentially retaking it.

I'm going to make a few tweaks to make that work smoother, particularly so the displayed work at the end will show right after the second attempt.  Also, I'm going to remove the individual-question reattempt links for this mode, to better match with the "only whole test can be reattempted" statement.

PR #44, opened 2016-07-18
Add button to allow JSON export of assessment questions by id 

kdv24:
# what
- work in progress!
- added a button to the ccexport page 
- button directs to assessmentexport.php
- currently var_dumps original array, then both var_dumps and echoes JSON and XML for given course > modules  > assessments > question ids 

# why
- to allow integration with Open Assessments via QTI

# test
- in course home page, click on export (under course items)
- on export page click on new button
image
- should show var dump of the current course with an array of the blocks, their assessments, and the individual assessment questions by id
- the results are also given in XML and JSON

# notes
This is a work in progress, but would love input on how to format, where to place the button, or any other insight you think would be helpful.

### DIFF
--- a/admin/assessmentExport.php
+++ b/admin/assessmentExport.php
@@ -1,0 +1,63 @@
+<?php
+require("../validate.php");
+if(isset($_GET['cid']))
+{
+  //courseid 
+  $cid = $_GET['cid'];
+  //Empty array that will return the blocks and all contents
+  $modules_array= array(); 
+  //returns all blocks based on given course id
+  $modules= returnCourseBlocks($cid);
+  
+  //finds all assessments for each block
+  foreach ($modules as $block) {
+      //creates an empty array to store names of all assessments in current block
+      $assessment_names_array= array(); 
+      //loops through an array of typeid(assessment typeids) in the block
+      foreach($block['items'] as $typeid) {
+        //finds names and ids of all assessments for the given block 
+        $query = "SELECT imas_assessments.name,imas_assessments.id FROM imas_assessments
+        join imas_items 
+        On imas_assessments.id = imas_items.typeid 
+        WHERE imas_items.id = ". $typeid .";" ;    
+        $result = mysql_query($query) or die("Query failed : " . mysql_error());
+        //adds the question ids to its specific assessment name
+        while($assessment = mysql_fetch_array($result)){
+          $AssessmentQuestionIds=returnAssessmentQuestionIds($assessment['id']);
+          $assessment_names_array[$assessment['name']]=$AssessmentQuestionIds;
+        }
+      }    
+      //adds the assessment names to the block array
+      $modules_array[$block['name']]=$assessment_names_array;
+    }
+
+//  exports the modules array as json
+    $outputjson = json_encode($modules_array);
+    header("Content-type: application/json");
+    header('Content-Disposition: attachment; filename="results.json"');
+    header('Content-Length: ' . strlen($outputjson));
+    echo $outputjson;
+}
+
+function returnCourseBlocks($cid){
+  $query = "SELECT itemorder FROM imas_courses WHERE id=" . $cid .";";
+  $result = mysql_query($query) or die("Query failed : " . mysql_error());
+  $items = unserialize(mysql_result($result,0,0));
+  return $items;
+}
+// takes an assessment id and returns all questions in a specific assessment
+function returnAssessmentQuestionIds($assessment_id){
+  $query ="SELECT imas_questions.questionsetid FROM `imas_questions` 
+  join imas_assessments 
+  ON imas_questions.assessmentid = imas_assessments.id
+  WHERE imas_assessments.id =".$assessment_id . ";" ;
+  $result = mysql_query($query) or die("Query failed : " . mysql_error());
+  $questionsetids = [];
+  while($row = mysql_fetch_array($result))
+  $rows[] = $row;
+  foreach($rows as $questionid) {
+    $questionsetids[]=$questionid['questionsetid'];
+  }
+  return $questionsetids;
+}
+?>
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+config.php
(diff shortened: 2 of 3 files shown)


kdv24:
@bracken & @AbeerKhakwani

admin/assessmentExport.php
```diff
@@ -0,0 +1,85 @@
...
+  echo "This is the JSON file: " . json_encode($modules_array);
+}
+function returnCourseBlocks($cid){
+  $query = "SELECT itemorder FROM imas_courses WHERE id=" . $cid .";";
+  $result = mysql_query($query) or die("Query failed : " . mysql_error());
+  $items = unserialize(mysql_result($result,0,0));
+  return $items;
+}
+// takes an assessment id and returns all questions in a specific assessment
+function returnAssessmentQuestionIds($assessment_id){
```
drlippman: This will probably work fine for your purposes, but just be aware this ignores cases where the assessment contains the same questionsetid multiple times.

drlippman:
Since this isn't a CC export, I'd rather the button for it not be on the ccexport page.  At the top of that page, you'll notice a "Export for another IMathAS system" link.  It would make sense to add a "Export OEA QTI" link up there next to it.  

It would probably make sense for that link to direct to a page that lists all the assessments for the course, with checkboxes for selecting them.  Then clicking Generate QTI would generate the output itself.

Separating this feature should also help prevent "what does that do?" kind of questions and confusion :)

If you need any other help with this, let me know.

admin/assessmentExport.php
```diff
@@ -18,41 +18,60 @@
...
-  $query = "SELECT imas_assessments.id,imas_assessments.name FROM `imas_assessments` 
-  join imas_courses 
-  ON imas_assessments.courseid = imas_courses.id
-  WHERE imas_courses.id = ".$cid .";" ;
+  $modules_array= array(); 
+  $assessment_names_array= array(); 
+  $module= loop1(1);
   
-  $result = mysql_query($query) or die("Query failed : " . mysql_error());
+  foreach ($module as $block) {
```
drlippman: Be careful here.  At any stage of the itemorder array, an element can either be an imas_items.id, or an array (indicating it is a block/module with a $block['items'] element. It looks like you're assuming every element of the array will be a block, and aren't accounting for blocks inside blocks.  You're also assuming that every item is an assessment, which is not true.  Oh, and you're using the imas_items.id as the assessment ID, which is not true.

Two options:
1) Just pull all assessments from imas_assessments using courseid as the filter, and don't worry about the itemorder at all, or

2) Recurse through itemorder, and pull from imas_items.  Maybe something like (untested):
````
$query = "SELECT id,typeid FROM imas_items WHERE courseid='$cid' AND itemtype='Assessment";
$result = mysql_query($query);
while ($row = mysql_fetch_assoc($result)) {
  $assessitems[$row['id']] = $row['typeid']; //this stores imas_items->assessmentid relations
}
function recurseitemorder($items,&$assessments) {
  foreach ($items as $item) {
    if (is_array($item)) {  //is a block - process contents
       recurseitemorder($item['item'], $assessments);
    } else if (isset($assessitems[$item])) { //if this item is an assessment, add to array
      $assessments[] = $assessitems[$item];
   }
}
$assessments = array();
recurseitemorder($itemorder, $assessments);
````

drlippman: If you end up building a page for selecting assessments to export, I'd recommend stealing code from /course/copyitems.php (lines 347-371 + 582-644), and you should be able to reuse the existing functions in /includes/copyiteminc.php.

If you have an item select page, then your processing/export script can just read the assessmentids from the POST.

drlippman:
Ahh shoot, I commented on an out of date diff.  Well the comments were still relevant, so hopefully you can see those :)

kdv24:
Great!  Thanks for the input. We'll fix it up.

kdv24:
@drlippman -  I kept running into problems with my json data disappearing when I used a temp file, so instead I just used unlink($my_file_I_created) to destroy the file after download.  Is that an acceptable way to deal with it?
Also, FYI, we're working on getting the initial data consumption to work across our apps, then will look at moving forward with your suggestion to have an intermediary page that lists all the assessments for the course, with checkboxes for selecting them.

drlippman:
Storing a file then destroying after download is fine.  I'd recommend putting it in /admin/imports/ since that directory is always set to be writable by the web server.

But I'm a little unclear why you need to write a file.  You're just writing it and then immediately reading it out again.  Why not just:
````
$outputjson = json_encode($modules_array);
header("Content-type: application/json");
header('Content-Disposition: attachment; filename="results.json"');
header('Content-Length: ' . strlen($outputjson));
echo $outputjson;
````

kdv24:
I had been thinking creating the file was necessary in case it needed to be used more than once.  However, since I ended up creating the attachment file that downloads, it becomes a user choice on the local environment whether to save or trash the file, thus negating my original reason for writing to a file (especially since I then realized I wanted to delete the file anyway!).  I know, dizzying logic.  ;)  So, thanks for the input and the cleaner solution.

kdv24:
@drlippman if this looks okay to you, can you accept this pull request and we'll keep moving forward on our end.  thx!

bracken:
@kdv24 If you think it's done now you can remove the "WIP" from the title.

drlippman:
There are several lingering issues that prevent me from auto-merging this, but I'll do a selective merge later today once I have time to cherry-pick and rewrite some of the block processing.

bracken:
kelly/abeer, you could squash and rebase on master to resolve any problems.

kdv24:
sure, I tried to rebase it last week and ran into issues pulling down the current master to rebase.  I'll take a look at it again.

kdv24:
okay, @drlippman I think we're good to go now.  All squashed, rebased, and happy.

drlippman:
Apologies for not doing a proper merge. I'm still not great with git subtleties around PRs, and I couldn't quite figure out how to do a merge with the changes I needed to make.

kdv24:
No problem. I faced my own challenges on the git cleanup!

PR #45, opened 2016-07-20
initial attempt for inter-question intro editing

githubapprentice:
This is an initial attempt with several bugs including what seems like a fundamental flaw with the numbering.  It doesn't yet have a UI.  Testing this requires manually editing imas_assessment.intro.

Since a text segment occupies a row in the table and needs a selectbox for reordering, I'm not sure if it's better if displayBefore indicates the question number or the starting row number.  This is an issue after reordering.  It wasn't clear what value should be sent to addquestionssave.php to put in imas_assessment.intro.  The row numbers no longer correspond to the question numbers the student sees, but I think the text segment rows need to be assigned a number for reordering.  (Or maybe this is a good reason to switch to drag-and-drop for reordering.)

If a text segment applies to questions 4 - 5 and question 5 is moved, something needs to happen but I'm not sure what.  I didn't handle that case.

It looks like the Assessment point total is counting the text segments.

There are a few other TODOs indicated in comments.

Sorry this isn't in better shape.  I'd like to explore options for the UI, but if you're up for getting the internals to work better, I'd like to hand that off.


drlippman:
I ended up picking-and-choosing from your PR, rather than merging.  I think I've got the basic framework working, with the exception of Embedded display, which still needs some work.

githubapprentice:
After making these changes to the UI, a bug surfaced when two text segments are at the end of an assignment.  I don't know all the use cases, but this doesn't seem like a useful case.  I'm happy to either prevent that case or work on solving the problem, but any tips on what might be causing the problem would be helpful.

This uses the approach of saving the changes when the text segment editor loses focus.  If that might be unclear to users, maybe a (fake) button would help at the expense of looking tidy.  Or I imagine there can be a Save button in the editor.

There was a problem with newlines in jsarr.  This code strips out newlines, but it there's a better way to escape them or a more robust way to JSON-encode the array, I'm happy to work on it.

Ideally, the background of the text segment in the question list would match the background used when it's displayed (e.g. white) rather than using the alternating row color.  What do you think?

There are new parameters for initeditor for added flexibility.  Hopefully this didn't break another other tinyMCE instances on other pages.

Any other suggestions are welcome.

drlippman:
I'll see if I can find a good workaround for the end-of-assignment text segment bug.

Interesting approach, using the inline edit mode of tinymce.  To be honest, I'm still trying to decide if I like it, since it causes the page to jerk and reflow so much, but that's going to happen unless the editor was done as a modal or greybox. It's probably fine, and I just need to get used to it :)

At the top of generateTable, add `tinymce.editors = [];` to clear out any previous editor instances.  This should prevent an error I see happening, I think caused by tiny trying to access the editor instance the function overwrites.

I am noticing that the blur action doesn't always seem to be triggering correctly.  Specifically, if text segment box is collapsed, I click on it, don't make any changes, and click out of it, the blur action doesn't trigger and the box doesn't collapse.  I think that's because your savetextseg action only does anything if the boxes are "dirty".  There should probably be an "else" on that block that collapses the cell if it hasn't been edited, calling `refreshTable()`.

I'm noticing clicking from one text segment to another doesn't work, probably because the whole table is being rewritten on the collapse/expand.  Probably not a big deal.

Small style thing: it would be nice if .textsegment.mce-edit-focus had 10px padding, so the text wasn't right up against the edge of the edit box.

drlippman:
Got the text segment at the end bug fixed.

drlippman:
Just a thought - what about not using blur?  My thinking is that in some cases, it might be nice to open up all the text segments at once, so you can see the whole flow of the text portions.

As you suggested, a "Save" or "Done" button could be used to save the text changes (if needed) and collapse the display.  Ideally there'd be a "you're going to lose your changes" notice if they trigger some other action before saving (or, other actions could trigger a save).

githubapprentice:
I agree it's likely someone would want to see all the text segments at once.  I guess there are two questions to consider: Should blur save the contents?  Should blur collapse the text?
A few options for collapse/expand:

1. Once a text segment has been edited, it doesn't collapse
2. Always show the full text segment
3. Use something like a [+] button to expand or collapse each text segment
4. Add a [+] button in the column header to expand all or collapse all

For saving, it could also be triggered by a timer.  Does save-on-blur seem reasonable at a minimum?  While saving, do we need to disable any other buttons/links?

Good point about the jerkiness.  How could we apply a style like 

    $(".textsegment").css("transition","height 0.5s");

to the editor or its container or to reflow in general?

Thanks for the other suggestions.  I'll work on those.

drlippman:
> Should blur save the contents
On one hand, it makes sense.  On the other hand, sometimes I'm editing text I make a mistake (accidentally delete stuff or something), and want a way to cancel.  Having it auto-save on blur (or via a timer) doesn't allow for a "discard changes" option.  I'm kind of leaning towards saying the text should not auto-save on blur, and there should be some clear "save" button.

Now, there could be an onbeforeunload event that checks for unsaved changes and asks the user if they want to save before exiting.  There should probably be one anyway to prevent exiting the page if a submitChanges action has been initiated but hasn't completed yet.

>Should blur collapse the text?
I'm thinking not.  I'm liking the idea of a separate toggle for expanding and collapsing, and for the collapse toggle to be separate from the "save" button.

githubapprentice:
Makes perfect sense to add Save, but I wish there were a way to avoid adding a button to that page.  At least it could be hidden when not editing.  A couple more questions to explore all the options:

1. If the editor's Undo button worked even after a save and if there were a "discard changes" button in the editor, what would you think about save-on-blur?

2. What do you think about having the save button/icon in the editor button bar?

I'll work on a separate expand/collapse toggle that doesn't look like an additional button.

drlippman:
> If the editor's Undo button worked even after a save and if there were a "discard changes" button in the editor, what would you think about save-on-blur?

The "undo" in tinymce does not always work reliably.  I'd prefer to avoid the save-on-blur.  However, if you created a copy of the original text when the editor was opened (and updated every time a hard save was made), and include a "discard changes" button that would reload that stashed copy of the original text, then maybe the save-on-blur would be OK.

> What do you think about having the save button/icon in the editor button bar?

I like this idea better, personally, so that I could save along the way easily.  I'd just want to make sure we can style it somehow to stand out (different color or something), otherwise I'm afraid people won't notice the button.

Gosh darn it's nice to have someone to talk through this kind of stuff with :)

githubapprentice:
There seems to be an issue with using tinymce's isDirty() function.  Since the editor contents were being re-initialized with the expanded contents of itemarray, it considered the contents dirty.  Using the isNotDirty property after that didn't seem to have a lasting effect.  So my current plan is to initialize the editor with the full text segment and use a style to show a collapsed version.

At the moment the Save button in the editor only saves the current editor, but it might be better to have it save all editors (if isDirty).  I guess there's a chance someone will have unsaved changes in one editor that they intend to discard, but then click save for another editor.  On the other hand, that doesn't seem as likely as someone clicking Save thinking it will save their changes in multiple editors.  So my plan is to have it save all.

Let me know if see any problems with either of these approaches or have any suggestions.

githubapprentice:
This uses styling to collapse text segments using a fade to show that there's more text.  The entire contents are loaded into the editor initially.  Unfortunately, this fade technique causes a portion of the div to not be clickable.  Also, if the text segment is just one line, it causes the bottom border to be faded.  It also collapses page titles if they are long, but maybe that looks too weird.  Is it better to simply not collapse page titles at all?
After a save, the tinymce button bar doesn't hide.  I'd like to fix that but haven't looked into it.
For some reason the Backspace key doesn't seem to work at all in these inline editors but does work on other pages.  It seems like something must be going out of it's way to do that.  Any clues what might cause that?
I added some class styles to default.css.  I hope that's okay.  I think I also need to add them to the high-contrast themes.  Not sure about other themes, they don't inherit from default.css do they?

drlippman:
Sorry, I've been otherwise occupied, and am just coming back to this.

> I added some class styles to default.css. I hope that's okay. 

If they're general styles, they should go in /imascore.css. If they're specific to this page only, they should probably go in the $placeinhead at the top of the page.

I have to say, I don't love the fade thing.  It seems really subtle, and if the first line is followed by a blank line, you really can't see anything at all.

Another note: if I click in a text segment and it expands, and I don't edit it, there's no way for me to collapse it again.  Oh, I see the [+] you put in the header now.  That's not very obvious, and seems to override the default expand-on-click behavior, which I found confusing.

> For some reason the Backspace key doesn't seem to work at all in these inline editors

That's because I try to catch the backspace key outside text fields, to prevent it from triggering the page-back action.  The inline tinymce fields aren't textarea or input fields, so the script is blocking the backspace.  That could be tweaked in general.js if needed.

So, as I'm guessing you are, I feel a little conflicted on the best way to proceed with the UI behavior here.  Right now it just still feels "clunky", though I'm the first to admit that a lot of the UI in IMathAS is clunky ;)

githubapprentice:
Let's figure out how to do this better.  Which of these sound best?

1. Instead of fading, shrink the font of the collapsed div for a thumbnail effect.  (Full or partial view.)
2. Add an "Edit >>" or "More>>"  button shown at bottom right of each collapsed div. Or use the notation from the blue Entry Tips box under an answer box.
3. Add a [+] link over each div like http://www.dynamicdrive.com/dynamicindex17/switchcontent2.htm
4. Collapse unedited text segments on mouseleave or focusout.
5. To prevent reflow, use something like "position: absolute" when the editor is activated for a dialog-like effect.
6. Create a dialog with the same location as the collapsed div but covering the following rows.  On Save, it's removed, showing the collapsed div.
7. Never collapse text segments.  WYSIWYG.
8. Post a few mock-ups on a MOM forum and ask for input and other suggestions.

Sounds like the location of style definitions and backspace issue are solvable.  Thanks for the tips.

drlippman:
I think I'm liking the idea of:

1. An explicit click-to-edit for the text segments, rather than doing the on-focus thing.  That could be a button at the bottom right of the collapsed div, a [+] button at the front of the text, or a separate edit link in the row (as there is now); I'm ok with any of those.
2. An explicit click-to-close/save button once editing, instead of doing onblur.  One button that does both, but only saves if there are changes.  Maybe a "Close" button that changes to "Save" if there are changes?  I'm ok giving up the "save but don't close" action.  If this button is in the editor toolbar, instead of using the default tinymce save icon I think it could be clearer to use a text-labeled button, in a distinct color to draw attention to it.
3. A beforeunload handler that checks for unsaved changes and gives a warning.

githubapprentice:
Here's an improved UI with an explicit click-to-edit.  I liked the idea of combining the close button with the save button, but after looking at it, thought it might be more intuitive if the close button is where the user clicked to expand it.  See what you think.  The beforeunload handler should be working but let me know if not.
A few caveats and questions;
1. When expanding all text segments at once, should any editor be active? If not, what button should be displayed in the individual editors: Edit or Collapse?
2. This uses the tinymce technique for icon-fonts.  Seems to be reasonably lean, but let me know if it's a problem.
3. If the user clicks in an editable div rather than the button, the editor activates before div fully expands.  In this case, if the page happens to be scrolled so the div is just below the top of the window, the button bar appears in the wrong place and requires another click or keypress to correct itself.  (There's also a dashed editor border that looks strange while it expands.)
4. Internally, tinymce hides the editor button bar when the input box looses focus.  I don't know if we could prevent that.
5. At some point when deleting entire paragraphs within the editor, there was some jerkiness which I can't reproduce.  There was also a slight jolt on the first time a box expands.  Let me know if see any of these.
6. The styles are now in addquestions.css
7. We need to do something about the Edit link in the Source column.
Let me know if you see any problems.  Happy to make more changes to make it easier to use.

drlippman:
Sorry I'm being slow getting back to you - I've been mired in the PDO conversion work.

I'm liking this better.  A couple things I'm noting:

1. While there's an beforeunload, there's no warning before other actions take place, like adding a next text segment, reordering, etc.

2. The expand/collapse icon isn't showing for me, just a faint outline around where it should be.  It looks like maybe the path for loading iconfonts/style.css is incorrect. Or maybe you forgot to push a file.  Can you just add those commands to the addquestions.css?

3. I think we could probably ditch the Edit link in the table row, and expand the textarea colspan across those additional columns.

4. I'm noticing line 59 seems to be throwing "cannot read property getContent of undefined" errors.

5. Super picky, but I think your expand animation could be a bit faster. It's nice to illustrate that the area is expanding, but probably 200 or 300ms would be sufficient.

6. You mentioned an "Expand all editors at once", but how would one do that?  I like the idea of adding that option (maybe a button after the +Text button that appears when there are at least one text item on the screen?  In that case, I don't think you need to apply focus to any of them automatically - the user can click into whatever they want to edit.

Super work on this.  You're making it much fancier than I would have ;)

githubapprentice:
This seems sooo trivial compared to PDO.  Thanks for your work on that.  2500 queries!

Sorry about the missing files.  The icons use the same tool as TinyMCE to create the css file.  It could be appended to addquestions.css but since it's auto-generated, it might be easier to maintain as a separate file.  There are also svg files etc.  Eventually they could be used elsewhere, but let me know the best place for the files.  Hopefully I pushed them this time.

These button tooltips are using the title attribute.  If it should use tipshow(), I'd be happy to change that.

I don't know if the Expand All is going to seem too disorienting on a small screen.

I had trouble finding where Backspace keydown is caught.  If you point me to it I'll take a look, or if it's easier to change it yourself, that'd be great.

I'm up for fine tuning.  Let me know if there's anything else.

drlippman:
Ok, more comments! :)

1. For now, maybe don't do the fancy styling on the add-text-button.  I think it's better for the styling to be consistent with the other buttons.

2. The Save All only seems to turn red when you move outside the text area.  It didn't make sense at first, but I guess I get the idea - call attention that they should save their changes.  Maybe make the color less glaring?  And adapt the :hover style to coordinate?

3. The button tooltips and expand-all both work fine to me. 

4. You have different warnings on the beforeunload and the table change.  Maybe make them a bit more consistent,  I like the latter for it's clarity about what needs to be saved.

5. Maybe remove the `background-color:#fff` from the .collapsed style.  I think it's find to let the .intro style show through, and it probably makes more sense since the expand/edit button it showing on top of the .intro style.

6. Should there be a "collapse all" option when there are unsaved changes?  It's probably ok, just asking.  If we're going to allow that, should we allow single-item collapsing without saving?

7. To fix the backspace issue, look in /javascript/general.js, line 758.  You'll want to add to the list `div[contenteditable='true']`

There are still parts of me that would like to be able to save one text area without saving & collapsing all, but I think that's ok...  

Now I need to start writing a converter to convert existing assessments into this format.

drlippman:
I just was playing around with converting assessments, and converted one that had really long blocks of text (it's a ebook with questions embedded).  I noticed a few things:

1. I would click on a text item, it would expand, then I wanted to collapse it, but I couldn't easily because the collapse icon was way at the bottom.  Maybe put the collapse icon at the top, or in the editing menu bar, or have it follow the user as they scroll.

2. When I collapsed the text item, I was left staring at a list of potential questions.  Maybe figure a way to adjust the scroll when the item is collapsed.  Maybe something like (this is a guess) where `this` is the collapse button:

````
//before collapsing
var initialdistfromtop = $(this).offset().top - $(this).scrollTop;
//after the collapsing
$(this).scrollTop($(this).offset().top+initialdistfromtop);
````

drlippman:
Another discovery:  The fitVids function from /javascript/general.js, which is used to help make video embeds scale down when they don't fit on the screen, is kicking in on with the inline tiny editor.  

Adding this to general.js, at line 712 seems to fix the problem by skipping the fitVids on the .textsegment elements:
`if ($this.closest(".textsegment").length>0) {return true;}`

drlippman:
Duh - just realized I should just make those changes to general.js in my version, and you can pull them in :) I'll do that right now.

githubapprentice:
Thanks for the changes to general.js and the converter!

I tried to address all the points above.  On the issue of saving just one editor, I don't know if you meant change the way submitChanges() works to not refresh the entire table.  I didn't take that on, and since only one save button is visible at a time, it seemed too easy for someone to click Save thinking it would save all the changes they made on the page and loose other changes.

I made this project harder than it needed to be by coding in a race condition.  I had the editor collapsing and adjusting the scroll while the button was following scroll.  The collapse would sometimes complete before the follow did, and the button would end up in weird places on the page. :)  Tracking it down stretched my debugging skills.

I did less regression testing on Embedded assignments if that helps to know.  Let me know if anything needs attention.

drlippman:
Looking good!  I'll probably make some small tweaks after I merge, but generally this is looking great.  Are you "done" enough that it's OK for me to merge, or are you still fiddling?

githubapprentice:
Tweaks are always welcome.  You can get them done orders-of-magnitude faster than I could.  I'm ready to move on to the next task: maybe buttons or calendar.

drlippman:
I vote calendar.  I had to restructure an online class the other day from 8wks to 10wks, and having a drag-and-drop calendar would have made it so much easier :)

drlippman:
Made a few tweaks.  See what you think.

githubapprentice:
Looks polished!  After the first time the Expand All button is used, it looks like they both do Collapse All.

drlippman:
Good catch. Got that fixed, and tweaked the CSS on collapsed textsegments so that header text at the top of the segment is shrunk down a bit so it's visible.  

I want to integrate the assessment converter somehow, then I think this is ready to merge into master!

drlippman:
FYI, found an issue with this - it breaks the add/remove question page when accessible (text-based entry) mode is used, which doesn't load tinymce.

Either we need to 1) force the load of tinymce, and assume it's accessible enough, or 2) rework it to only use textboxes.  Unfortunately since the save functionality was built into tiny, I think doing the latter might be too difficult.

githubapprentice:
Sorry I didn't think of that.  If HTML supports an editable DIV, could we just add that attribute and a mechanism to trigger the save?

PR #52, opened 2016-08-28
JSXGraph Library

gksander:
JSXGraph Assessment Library. Still unsure on how to best hook the JS
code produced into the question so that it is non-blocking

### DIFF
--- a/assessment/libs/JSXG.html
+++ b/assessment/libs/JSXG.html
@@ -1,0 +1,284 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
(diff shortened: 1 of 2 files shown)


gksander:
(This is my first ever git/github experience - so I apologize in advance for my noob-ness.)

drlippman:
Sorry - I got caught up in some other stuff and totally forgot about this. I'll take a look again and see if I can come up with a good solution for the JS hook.

gksander:
No worries - I also got caught up in other stuff, and just now thought about this this morning.

The JS code for the hook is at the end of the JSXG_createAxes() (line 166) and JSXG_createPolarAxes() (line 1043)

drlippman:
I rewrote the script call to first look and see if it's already been loaded, and that seems to be working.  I also changed the call to makeBoard to use the system's `initstack` script loading stack, which prevents race conditions (where the board was being initialized before MathJax was ready).  Going to merge this in.

PR #53, opened 2016-09-02
JSXG Geometry 1

gksander:
First round of geometry additions: points, segments, lines, rays, angles, circles, polygons, gliders.

Allowed for javascript object reference, so elements can build off each other. Allow for overriding element attributes using JSON string (JSXGraph style).


drlippman:
Sorry it took me so long to look at this.

PR #55, opened 2016-09-18
Style.consistent calendar colors using less

githubapprentice:
Sorry some of these diffs are overwhelming.  I should have done the first commit in two steps.  The third commit is a good starting point.
I wasn't sure if the .less files should be in the same directory as the .css but this uses a subdirectory.  What would be the best structure?
The less/readme.txt explains to a developer how to compile the Less files.  The documentation for installation was cryptically general, but it's actually pretty straight-forward.
In terms of actual style changes, this only changes the calendar headers to match the leftcontent colors and moves a few calendar styles from imascore to the theme files.  If it helps, I'll send you a patch for the otbs_fw.css file.  I'd like to also change the today color if the concept is okay.  I don't really have the big picture of how other CSS files (or styles in PHP) might overlap etc.
Let me know if you have any suggestions, especially for the overall concept.


githubapprentice:
This separates the changes made to imascore.css into a different commit than renaming the file.  Hopefully clearer what changed.  Also merges changes to span.calitem in imascore.

drlippman:
I'm going to go ahead and close this PR.  While I appreciate the idea of what you're trying to do here and the effort you put in, I think the style system probably needs a more comprehensive overhaul from the ground up.

PR #58, opened 2016-11-09
Add info about rounding to Entry Tips

githubapprentice:
This shows a rounding example in the Entry Tips and includes a message about decimal places or significant digits.

In some places the phrase "significant figures" is used.  Although that's a standard term, I wonder if changing all of them to "significant digits" would be clearer to students who haven't been taught either phrase.

Due to problems with my development server, I wasn't able to test the short version of the Entry Tips and testing was limited in general.

Do I need to do anything about internationalization?

### DIFF
--- a/assessment/displayq2.php
+++ b/assessment/displayq2.php
@@ -945,19 +945,21 @@ function makeanswerbox($anstype, $qn, $la, $options,$multi,$colorbox='') {
 			$rightb = '';
 		}
 		if (in_array('list',$ansformats) || in_array('exactlist',$ansformats) ||  in_array('orderedlist',$ansformats)) {
-			$tip = _('Enter your answer as a list of whole or decimal numbers separated with commas: Examples: -4, 3, 2.5') . "<br/>";
+			$tip = _('Enter your answer as a list of whole or decimal numbers separated with commas: Examples: -4, 3, 2.5234') . "<br/>";
 			$shorttip = _('Enter a list of whole or decimal numbers');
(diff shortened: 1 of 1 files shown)


drlippman:
Sorry I haven't done anything with this yet.  I have a potentially conflicting accessibility change I might need to make first.

drlippman:
So looking at this again, my concern is that saying "4 significant digits" isn't necessarily true, if the user has changed the $abstolerance or $reltolerance.  I guess we could try to write a function to infer number of decimal values / sigfigs based on those, but I feel like sometimes I set those looser than what I tell students, which could lead to confusion.

I do like the idea of giving the example decimals as ones with 4 decimal places.  I'll go ahead and make that change.

githubapprentice:
That seems fine to me, but is it worth adding else cases to handle $abstolerance or $reltolerance so the entry tips automatically have instructions about rounding?

drlippman:
I'm a bit concerned that using $abs/reltolerance to auto-create entry tip rounding instructions might be problematic in some cases. As I mentioned, sometimes I set those looser than what I tell students, which could lead to confusion if they see conflicting rounding instructions.  But maybe I'm the only one who does that :) 

If we did, $reltolerance would probably lead to a "give your answer accurate to at least __ significant figures".  $abstolerance should probably be "give your answer accurate to at least __ decimal places".

But what would $abstolerance = .005 translate to?  Accurate to at least 2 places? 3?

drlippman:
I daresay the _right_ answer is probably that people should use $reqdecimals or $reqsigfigs instead of $abs/reltolerance, but obviously that doesn't help with all the existing questions.

PR #61, opened 2017-01-05
Jsxgpatch

gksander:
JSXGraph Library: make JS board variables global so that they can be used in JS code in question.


XYZPatrick:
@codex review

PR #62, opened 2017-01-05
Codemirrortweak

gksander:
Added line numbers and HTML tag auto-closing to Question Text editor when in markup mode. (These addons are packaged in codemirror-compressed.js, but not enabled in config.)


drlippman:
I remember I intentionally didn't include line numbers in the qtext field because they seemed pretty unnecessary for that field.  What's your motivation for adding them?

I seem to remember I intentionally left out auto-closing as well, but can't remember why.  I'll have to play with that a bit.

gksander:
I'm using JSXGraph quite a bit in questions, adding JS to the question text area. Sometimes my question text will get up to around 100 lines. I like having the numbers there for when I'm scrolling through that many lines of code. 100 lines isn't that much, so I can definitely live without.

As for the auto-closing tags: after using that feature (or something similar, e.g. Emmet) in a desktop editor, it feels awkward and exhausting having to close out all of my tags, and hitting the return button. I've probably just gotten lazy, so I could live without the auto-closing. I just thought it was a convenient feature, and wasn't sure if you left it out intentionally or not.

gksander:
After using the tag auto-closing for a day on my local install, I see why you didn't enable it. CodeMirror's auto-closing isn't very "intelligent" and ends up being somewhat annoying. You can disregard that request.

PR #64, opened 2017-01-06
Ajax qEditor

gksander:
A first stab at making a quick-save option in the question editor. Everything seems to work in my local dev. Only thing I couldn't figure out was how to bind the quickSaveQuestion method to tinyMCE keydown event.


gksander:
After writing some questions with this:

If the problem isn't *first* saved with a full Save or Save and Test, it won't preview correctly and it will continue to make duplicates of the question.

drlippman:
I'd suggest refactoring/rewriting this in a couple ways so that it can always use AJAX save.

1) I don't like the idea of forcing non-AJAX saves when there are file uploads.  How about instead you [test for FormData](http://stackoverflow.com/questions/4006520/using-html5-file-uploads-with-ajax-and-jquery) and [use it](http://portfolio.planetjon.ca/2014/01/26/submit-file-input-via-ajax-jquery-easy-way/) to submit the form, including files.  If the browser doesn't support FormData, then just disable AJAX-save altogether.

2) To handle the first save case, I'd suggest that you go ahead and submit it via AJAX, and have the save function return the new question ID.  Then you can use that to rewrite the form "action" URL and the URL you post future AJAX saves to.  If the browser supports it, you can use window.history.replaceState to also rewrite the URL that shows in the address bar.

Also, as a stylistic thing, please use full <?php tags, not the <? short tags.  The latter is generally discouraged, as support for them can be disabled in the php.ini.

gksander:
Great suggestions -- I attempted to incorporate them. Thanks for the nudge in the right direction.

1. I tested for FormData, and disabled quick-save if the object isn't detected.
2. I handled first save case essentially as you suggested. I built the form action early in the php script, and sent that as part of the AJAX response. Then used that to set the query string and form action.
3. To handle images, I sent the $images array back with the AJAX response, and then used it in the JS script to generate the image list. 
4. I removed php short tags.

gksander:
I did some battle testing with this on localhost and think I've worked out all of the kinks. The quicksave option feels really nice (to me) when authoring questions.

drlippman:
Glad to hear it's been working well. I'm hopeful to find time to review and pull this soon.

drlippman:
Thank you for all your work on this, and sorry it took me so long to review it and get it merged it.  I merged the changes in commit 71e19199de9232b75042e63c741c4380346ca9c3 with only a couple minor modifications to address the listing of extrefs; I'm not sure why github didn't recognize it as a merge - perhaps because I made some additional changes and then merged to master as a squashed merge.

PR #67, opened 2017-02-09
forminlinebutton Patch

gksander:
Macro function forminlinebutton is re-labeling/re-numbering incorrectly in embedded assessments. This is an attempt at patching that by adding an option fifth parameter to the function used to label.

-------------------------

I think that this could be done more elegantly, perhaps by slightly modifying the assessbackgsubmit funciton in AMHelpers.js. I now see how the assessment .js files are being minified, so I'll take a run at trying to do this via assessbackgsubmit.


drlippman:
Hmm.  My only concern would be that any existing uses of forminlinebutton won't have an $id specified, so this would fix the reload issue for those questions.

Maybe a better strategy would be eliminating the use of an ID altogether, and rewrite the toggleinlinebtn function.  I think we could replace the `id="inlinebtn'.$r.'"` with `class="inlinebtncontent"` then in toggleinlinebtn have it look for the immediate sibling to the item clicked with class inlinebtncontent, and toggle the display on that.

I can probably throw that together pretty quickly... let me try it.

drlippman:
Nevermind.  We need an id to add aria-controls to that, which is needed for accessibility. I think I can use uniqid() to generate an ID where we won't have to worry about conflicts.

PR #71, opened 2017-03-03
Conditional .ansred .ansgrn block

gksander:
.ansgrn and .ansred use display:inline-block, which works well for input elements. “Squeezes” things when used in conditional questions, so added inline display:block for that case

### DIFF
--- a/assessment/displayq2.php
+++ b/assessment/displayq2.php
@@ -241,7 +241,7 @@ function displayq($qnidx,$qidx,$seed,$doshowans,$showhints,$attemptn,$returnqtxt
 	if ($qdata['qtype']=='conditional') {
 		$qcol = isset($qcolors[0])?(is_numeric($qcolors[0])?rawscoretocolor($qcolors[0],1):$qcolors[0]):'';
 		if ($qcol!='') {
-			$toevalqtxt = '<div class=\\"'.$qcol.'\\">'.$toevalqtxt.str_replace('"','\\"',getcolormark($qcol)).'</div>';
+			$toevalqtxt = '<div class=\\"'.$qcol.'\\" style=\\"display:block;\\">'.$toevalqtxt.str_replace('"','\\"',getcolormark($qcol)).'</div>';
 		}
 		if (!isset($showanswer)) {
 			$showanswer = _('Answers may vary');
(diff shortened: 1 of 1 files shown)


assessment/displayq2.php
```diff
@@ -2756,8 +2756,8 @@ function makeanswerbox($anstype, $qn, $la, $options,$multi,$colorbox='') {
 						if (count($answerformat)>1 && $answerformat[1]=='opendot') { $out .= 'class="sel" '; $def = 2;}
 						$out .= ' alt="Open dot"/>';
 					}
-
```
gksander: I'm a git noob, so I'm not sure why I'm generating line additions/deletions when committing... Only intended to change line 244.

drlippman:
Can you explain better what you mean by "squeezes" things, and why it's a bad thing?  I used inline-block exactly so the box would "fit" better to the content, rather than being full-width.

gksander:
"Squeeze" probably didn't convey the meaning I intended. The content container will resize when display changes from block to inline-block -- but this change in size isn't noticeable in a lot of cases. I was noticing it when the div.ansgrn element didn't fill the the entire width of the question container -- it felt like the content was being "squeezed" to the left, more so than the original display of the question. This is non-intrusive, so I guess this isn't a bad thing.

One simple (and contrived) example that portrays what I mean... in question text of a conditional question (with $anstypes = ['numfunc', numfunc'];):

`<p>Define 2 equivalent functions.</p>
<div style='text-align:center'>
  <p>`f(x)=` $answerbox[0]</p>
  <p>`g(x)=` $answerbox[1]</p>
</div>`

I understand your rationale to have the div.ansgrn fit content, so this request can be closed.

drlippman:
Ahh.  I see what you mean.  Though I'd argue wrapping answerboxes in a `text-align:center` div is a pretty bad idea.   It looks really weird to have your entry box start crawling to the left as you enter stuff and the live preview shows.

I personally really don't like having the ansgrn box fill out the full width of the question div, so am afraid I'm not in favor of this PR.

drlippman:
I may have to reconsider this, as the inline-block could potentially also mess up the display of full-width tables and other such things.  Hmm...

PR #75, opened 2017-03-26
Translation and theme fixes

sbaecker:


drlippman:
Whoops - not sure when I messed up the $theme

PR #77, opened 2017-04-14
QuickSave Button for Detailed Solution

gksander:
Add Quick Save and Test button to Detailed Solution section for
consistency.


drlippman:
Thanks!

PR #91, opened 2017-07-19
Wording changes in addassesment.php.

kittybiscuits:


drlippman:
Thanks for making the PR, but I'm not a fan of this change unless the language is changed elsewhere to be consistent.  I think some broader consideration needs to go into the language and format of the date options across the system.

PR #98, opened 2017-10-18
VideoCued Aspect Ratio

gksander:
Add option for video aspect ratio in video cued assessments.


gksander:
Added option for video aspect ratio. Defaults to 16:9. Note: in the viddata field in imas_assessments, the array now looks like [vidid, vidar, ....] instead of [vidid, ...]. 

The only issue I'm running into is... on /course/addquestions.php, when removing a question, the aspect ratio gets wiped. I think this has something to do with the code around line 90, but I haven't quite made sense of what's going on there. (This doesn't happen when a question is added, however.)

I'll look into this problem in the next day or two and add another commit if I get it figured out. If you can quickly spot how to patch that, feel free to modify/merge.

drlippman:
Yes, I think the `vidar` getting wiped is probably happening where you suggested, and also in `addquestionssave.php`, which is used when removing questions.

It might require less code changes to revise where/how you're storing the aspect ratio.  Rather than adding it as an additional element in the array, causing everything to shift and you having to adjust for that, how about changing the first element from just a vidid string to an array?

It's currently `[vidid,firstviddata,...]`, and you're using `[vidid, vidar, firstviddata,...]`, but how about instead you do ``[[vidid, vidar], firstviddata,...]`?

With that, the existing indexing used in addquestions/addquestionssave will still work.  In `showtest.php` you could keep the work you've already done for the most part, but after doing `$vidid = array_shift($viddata);` the next line could determine whether the first element is just the vidid or an array also including the aspect ratio:

```
if (is_array($vidid)) { 
  list($vidid,$vidar) = $vidid;
} else {
  $vidar = "16:9";
}
```

gksander:
Great idea! I just did that, and made the commit. Seems to be working the way it is supposed to now.

drlippman:
If I may make a request for a small change, on line 44 would you change `array_unshift($data, [$vidid, $vidar]);` to `array_unshift($data, array($vidid, $vidar]));`?  The shorthand notation was added in PHP 5.4, and I'm not sure everyone is running that yet.

gksander:
Done. (I didn't realize [..., ...] was a relatively new syntax -- good to know!)

drlippman:
I just made a new branch based on your work, and also adding the responsive menu for video cued.  Please give it a test run and see if everything seems to be working.

drlippman:
Closing, since #99 replaces this.

PR #99, opened 2017-10-19
Videocued improvements

drlippman:
Building off @gksander's PR #98, this fixes a small bug in the aspect ratio, adds an onchange for aspect ratio on the editing page, and adds a responsive fly-in menu for video cued on smaller screens.


gksander:
Just tested this out -- everything looks like it's working on my end. I like the fly in sidebar.

One minor suggestion: adding a `cursor: pointer` to the hamburger, and maybe a `:hover{background-color: ...}`, like on the hamburger on the course home page.

drlippman:
Good idea.  Just the cursor, and went ahead and switched out the image menu icon with a CSS one so I could color it onhover; I thought that might look better than a background color.

PR #100, opened 2017-10-25
Add Emmet to CodeMirror

gksander:
Add emmet CodeMirror from https://github.com/emmetio/codemirror


gksander:
What are your thoughts on adding Emmet to the Question Text HTML editor? Coming from Sublime or VS, it feels very natural to have this extension. The CodeMirror tag completion extension is clunky and annoying, but this plugin is actually from Emmet and feels a lot nicer (and triggered on Tab, so it doesn't get in the way if you don't want it to). For example, `ol[style="list-style:lower-alpha"]>(li>p*2)*3` feels a lot nicer on the fingers then typing out all of the markup.

If you don't feel this is worth having in the question authoring page, feel free to disregard and close this PR.

drlippman:
Thanks for the suggestion.  Strikes me as pretty heavyweight (~1MB of script) for the number of people who would probably ever use it.  It seems like toggling on the tinymce editor and using the WYSIWYG editor is probably an easier solution in most cases.

gksander:
I agree -- 1MB is heavy for the number of people who would use it. The only issue with the tinymce editor is it sort of mangles the markup. It removes all indentation, so going back into HTML mode would get rid of all prior indentation.

I just synced up another commit where the script is loaded asynchronously on a button click. What do you think about that? It would allow users to use it if they please, but wouldn't load that script each time moddataset.php was loaded. (I just threw the button in there -- it would definitely need a better home. I was just prototyping. This could be just a small link hidden away somewhere unobtrusive.)

drlippman:
If that's the concern, it might be more generally useful to try to figure a way to prettyprint the HTML in Codemirror.  It appears there [once was](http://codemirror.net/2/demo/formatting.html) a formatting plugin for Codemirror but apparently it's no longer supported in the current version.  It looks like there (may be a way)[https://stackoverflow.com/questions/25109809/codemirror-auto-format-after-setvalue] to recreate it.

gksander:
Ah, yes - I agree! That would be more useful, because there are surely other reasons to toggle over to TinyMCE and back to Codemirror (image uploads, etc.).

drlippman:
Actually, it looks like all that's required to get most of the autoindenting to work is adding 
```
for (var i=0;i<qEditor.lineCount();i++) { qEditor.indentLine(i); }
```
in the `setupQtextEditor` function after the `fromTextArea` call.

It doesn't format `<br/>` tags well, but things like tables format nicely.

drlippman:
Part of the issue is that in converting from textarea to tinymce, the system converts double newlines into `<br/><br/>` to preserve the formatting (that's the same thing that happens when a question is rendered).

One option would be to convert double `<br/>` back to double newlines when switching from tinymce to textarea, but that's not idea since tiny wraps the text in a `<p>` tag...

drlippman:
Oh this might work.  Wrapping the double `<br/>`s in newlines, so
image
when converted to tinymce and back would look like
image

drlippman:
I added the auto-indent, so I'm going to close this PR.

PR #112, opened 2018-02-01
Update tinymce to 4.7.6

drlippman:
This could use some testing


drlippman:
Merged in separately.

PR #118, opened 2018-03-09
Wrap variable values in parentheses

gksander:
Negative slider values were killing applets. Wrapped values in
parentheses to fix this.


githubapprentice:
Thank you!

PR #128, opened 2018-06-15
Add closing divs to new messages/forums gadgets

gksander:
Missing `</div>` tags in  "New messages" and "New forum posts" gadgets on homepage. When no new messages, forum post gadget gets absorbed into message gadget.


drlippman:
Thanks!

PR #147, opened 2019-08-30
Bug fix: Typo in variable used in LTI queue processing callback.

kittysnacks:
Also log DELETEs on the imas_ltiqueue table.
This happens after a successful LTI grade return.


sketchings:
Only caused issues if LTI processing did not happen the first time

kittysnacks:
Updated PR to remove logging.

PR #150, opened 2019-11-07
Fix for wrong library order in addquestions.php

zjevik:
Libraries displayed in the Add/Remove Questions page are sorted by the library id and not how they display to the user in libtree2.php iframe. This commit fixes the issue. Detailed description of the problem is below.

Example: 
Let's say that we create two child libraries:
- 01 first
- 02 second

and now we rename them:
- 01 first -> 02 first
- 02 second -> 01 second

Assuming the parent library has Alphabetical sorting. Open the Add/Remove Questions page. We will see the libraries (in libtree2.php iframe) in the order:
- 01 second
- 02 first

But if you select them and click on Search to display all questions we will see that the libraries will be displayed in the order (sorted by the library id):
- 02 first
- 01 second

### DIFF
--- a/course/addquestions.php
+++ b/course/addquestions.php
@@ -777,7 +777,7 @@ if (!(isset($teacherid))) { // loaded by a NON-teacher
 				$lnamesarr[0] = "Unassigned";
 				$libsortorder[0] = 0;
 			}
-			$stm = $DBH->query("SELECT name,id,sortorder FROM imas_libraries WHERE id IN ($llist)");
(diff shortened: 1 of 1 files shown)


drlippman:
Nice solution to this.  I've never used this particular ORDER BY syntax before.

zjevik:
> Nice solution to this. I've never used this particular ORDER BY syntax before.

Neither did I know about this. It was quite intriguing to find a solution to this problem without rewriting a lot of the SQL. I'm happy the solution was so elegant.

PR #154, opened 2020-01-27
Translated forms.php and newusercommon.php

ingodahn:
Hi, David,
I extended the German translations to forms.php and newusercommon.php which required changes in the files. Please, pull.
Best
Ingo


drlippman:
Thanks!

PR #155, opened 2020-01-29
Use empty() instead of count() when checking for errors.

kittysnacks:
It's possible $errors is not always an array. Using `empty()`
allows us to check for false, nulls, empty strings, etc.

OHM-318: High error rate in scorequestion.php
(cherry picked from commit 92837c57c91d44ff7ef9dc9473a6e400f95a95da)


drlippman:
This looks fine, though I wonder why $errors would be coming back as something other than string.

PR #161, opened 2020-02-24
Feature de.addquestions

ingodahn:
Translated course/addquestions.php and javascript/addqsort.js. Button text in value attribute not translated yet


ingodahn:
Merged into feature_de

PR #166, opened 2020-04-13
added conditional call to alreadyLoggededIn function in validate hook

xyztextbooks:
Hi, David.

Currently, when $CFG['hooks']['validate'] is defined, the onLogin() function is called upon user login.

We need another function (alreadyLoggededIn) to be called in cases where the user is already logged in (this change provides an entry point for our student paywall code).

-steve

P.S.: This is my first attempt at a Pull Request, so I'm hopping I followed the correct procedure.


drlippman:
Your first attempt at a pull request was great :)

Do you want to pass the userid as an input to the function?  Or are you figuring that out another way?

xyztextbooks:
Thanks David!

Stephen made this request so I sent the email to him, to get back to you.

P

****************************
Patrick McKeague
XYZ Homework
805.441.7596

xyztextbooks:
I'm merely using $_SESSION['userid'] in my alreadyLoggededIn function, so I son't think we need to pass vars to it. Of course, adding a userid param would do no harm.

drlippman:
I think for the sake of isolation of concerns, it'd be nice to pass the $userid as an input to this hook, so that the hook function doesn't need to worry about how the userid can be looked up.

MathTV:
Thank you! That makes sense.

PR #167, opened 2020-04-29
ohm-571 default to new assessment player when copying from old player

sketchings:
When an LTI teacher user creates a new course via LTI using one of the copy options (create a copy from the original course or create a copy from a copy):

- the new course should be by default in the new player format
- if copying from a source course in old format, show option to opt out and use old player
- if copying from a source course in new format, do not show assess player option


drlippman:
Thanks for the PR, but this is fiddlier than I would like.

PR #168, opened 2020-05-05
Fixed: Unable to save assessment if groupmax < 2.

kittysnacks:


drlippman:
This issue should have already been addressed by line 73 in `addassessment2form.php`:

`'groupmax' => ($line['groupmax'] > 2) ? $line['groupmax'] : 6,`

There shouldn't be need to adjust it twice.  Though I am noticing it should have been `>1` in that line rather than `>2`, so maybe that's the issue.

drlippman:
This change shouldn't be necessary, but no harm in addressing it twice.

PR #170, opened 2020-06-04
Added vmsetupnumbertiles_ten_one

meachamw:
A request was made for the numbertiles vm without the 100's block.  Added a new function called vmsetupnumbertiles_ten_one.  This vm will use the same "vmnumbertilesgetcount" with the 100's block always being zero.
Will be doing a pull request to add numbertiles_ten_one.html to the library.


drlippman:
It looks like you still need to add the new function to the $allowedmacros push near the top of the file.

PR #173, opened 2020-06-24
Add support for displaying assessment name in calendar instead of caltag

xyztextbooks:
Reasoning:
- A number of our instructors use short assessment names (e.g., "quiz 1", "2.1"), and they have grown tired of retyping the assessment name into each assessment's caltag field.
- The suggested modifications enable display of assessment names with minimal code changes, i.e.: selecting the new "Use Assessment Name" radio button when editing an assessment merely sets the caltag field to "use_name". The calendardisp replaces caltag with name when caltag == 'use_name'.
- see attached screenshots for preview of changes.

Changes:
- added 'use text and 'use assessment name' radio buttons for caltag field in assessment options
- added 'use text and 'use assessment name' radio buttons for caltag field in masschange forms
- modified calendardisp to use assessment name if caltag = 'use_name'
- added explanation of $CFG['AMS']['caltag'] = 'use_name' to readme.md

[screenshots.pdf](https://github.com/drlippman/IMathAS/files/4828309/screenshots.pdf)


drlippman:
I generally like this idea.  Would it make sense to truncate the name on display if it's really long?  Or just allow really long names that will wrap and make the calendar look crazy? :)

xyztextbooks:
Yes truncate if too long:-)

Thanks!

P

xyztextbooks:
@drlippman Which truncation strategy do you prefer:

- use php substr function to shorten string and add ellipses
PROS: quick, easy, does not complicate front end code
CONS: not responsive to screen size

- use css to set overflow properties (e.g.: overflow: hidden; white-space: nowrap; text-overflow: ellipsis;) for the span.calitemtitle
PROS: better responsiveness to screen size
CONS: a bit more work, adds a little complexity to front-end code

-steve

drlippman:
While I like the idea of using css, that does require nowrap, which means calendar tags can't wrap at all.  Right now it's common for somewhat long tags to wrap to a couple lines and it doesn't look too bad, so I'm hesitant to restrict the display to a single line.  But if you can find a way to do that that looks OK, that's fine. Otherwise a basic substr should be sufficient.

xyztextbooks:
After a bit of reading, it looks like a clean css method will not exist until browsers consistently support a "line-clamp" property.

So, I think the options are:

- hide text that exceeds some max_lines_for_calitems that we choose (ellipses will NOT appear at the end of line 2), e.g., to show only 2 lines:
`span.calitem {
  overflow:hidden;
  line-height: 1.1rem
  max-height: 2.2rem
}`

- use PHP substr to set max string length.

I like the PHP substr option for now.

drlippman:
Sorry I've been so slow.  Yes, substr is probably the best option for now.

xyztextbooks:
OK. I modified calendardisp.php to truncate names longer than 40 characters:

`
	// if caltag == 'use_name', display the assessment name
	if ($row['caltag'] == 'use_name') {
		// truncate name, if needed
		$row['caltag'] = strlen($row['name']) > 40 ? substr($row['name'], 0, 40) . '...' : $row['name'];
	}

`

drlippman:
Merged manually, with a few changes.  Not sure why github isn't recognizing it as a merge.

PR #180, opened 2020-07-22
add support for prepending html content to chguserinfo form

xyztextbooks:
We need to display a bit of extra info (activation expiry for students, etc), and the chguserinfo form is a good place for it. The data I'm planning to insert will be read-only.

### DIFF
--- a/forms.php
+++ b/forms.php
@@ -16,6 +16,11 @@ if ($_GET['action']!="newuser" && $_GET['action']!="resetpw" && $_GET['action']!
 	$coursetheme = $defaultcoursetheme;
 }
 
+//Look to see if a hook file is defined, and include if it is
+if (isset($CFG['hooks']['forms'])) {
+	require($CFG['hooks']['forms']);
+}
+
 require("includes/htmlutil.php");
 
 if (isset($_GET['greybox'])) {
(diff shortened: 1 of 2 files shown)


xyztextbooks:
Here's a screenshot of what we're inserting.
Screen Shot 2020-07-22 at 5 13 08 PM

drlippman:
This is OK, but may I request you implement this in a slightly different way, using the "hooks" approach I've been implementing elsewhere in the system.   

There's a description of the existing hooks in `hooks.md` 

You can see an example of this approach in `/admin/actions.php`.  At the top of the file we look to see if `$CFG['hooks']['admin/actions']` is defined.  It should be defined to be the path to the hook file to include.  If the $CFG hook is defined, we include it.  The hook file itself defines functions.  Then in the code, where the hook would apply, we check to see if the function is defined and call it if it is.  For this chguserinfo change, you could have a hook function called `chguserinfo_extras` and where you want it to show you could do:
````
if (function_exists('chguserinfo_extras')) { 
  chguserinfo_extras();
}
````
And that function can output whatever you want to show.  

I'm suggesting this approach largely because we're using it elsewhere to add hooks for extending functionality, so it would be nice for others to be consistent.  It's also nice because you can include multiple functions in one hook file if you want.

xyztextbooks:
I started fresh and followed the standard hook strategy:
- added include of hook specified by $CFG['hooks']['forms']
- added call to chguserinfoExtras function

drlippman:
Looks great, thanks!

PR #183, opened 2020-07-29
Added a mode function to the Stats macro.  Also updated/formatted the Stats help.

duragauge:
The mode function accepts an array of numbers or strings and outputs a list which can be used directly for $answer.

Capture


drlippman:
Thank you for your work on this, but ultimately I decided to implement the mode function a different way using native PHP functions for doing the frequency count, which ended up being 50x more efficient.   Sorry I didn't give you a chance to do it, but I didn't want to ask you redo it without knowing if the native functions would even work and actually be more efficient.  I also ended up adding a forceonemode function that will revise the dataset to only have one mode.

duragauge:
No worries, thanks for the update!

PR #185, opened 2020-08-30
Added German translations

ingodahn:
Besides language files there are changes to the following source files in course/:

- course/autoexcuse.php
- course/categorize.php
- course/addquestions.php
- course/modquestiongrid2.php


drlippman:
Thanks, but I'm going to have to selectively manually merge this, since you have other changes in this PR other than the new translations.  I'll get to it when I can.

PR #188, opened 2020-09-11
adds capitalized pronouns to randnamewpronouns() macro

nick-chura:
I've seen folks request this a few times, so I thought it might be worth addressing.

**Code example:**
$name is my friend, and $heshe loves cats more than I do. $Hisher cat's name is Freddy. $Heshe has white fur with black spots. The cat, I mean...not $name!

**Output:**
Jake is my friend, and he loves cats more than I do. His cat's name is Freddy. He has white fur with black spots. The cat, I mean...not Jake!

### DIFF
--- a/assessment/macros.php
+++ b/assessment/macros.php
@@ -2420,15 +2420,15 @@ function randnamewpronouns($g=2) {
   
   if ($g==2) {
   	if ($gender==0) { //male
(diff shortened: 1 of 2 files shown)


drlippman:
I appreciate the idea and effort, but using `ucfirst` to capitalize manually seems simple enough.  This feels like it overly complicates things.

nick-chura:
Yeah -- I figured it was a bit of a long shot and a little complicated.

PR #190, opened 2020-09-22
A working unit-checker

nick-chura:
Hi David -

Please look this over for functionality -- I'm sure there are hidden issues I haven't found while testing it. I made it an option inside of Number-type questions, with $answerformat="units".

It seems to work with $requiretimes.

Things I don't know how to do:

1. Change the answerbox tip to accommodate "units" in the answerformats. The long tip should probably say something like

"Enter your answer in the form [decimal number] [unit]^[power] [unit]^[power] ... / [unit]^[power] [unit]^[power] ..."

and the short tip might say "enter a product or quotient of decimal numbers and units with exponents"

2. How to keep the student answer displayed in the answerbox after clicking "submit".

3. How to provide feedback on student answer formats.


nick-chura:
Changed the answer checker to use defaultreltol instead of ==. I know this should be changed to involve the other options (abstolerance, sigfigs).

nick-chura:
You can now type things like 5 microseconds (parsed as 5.0E-6 sec) or 5 microdeciseconds (parsed as 5.0E-7 sec) or 5 microdeciseconds^2 (parsed as 5.0E-14 sec^2). [Not that we expect anyone to do this...]

drlippman:
This has been merged manually.

PR #192, opened 2020-09-29
New Finance Library

ahtavangar:
Hi David,

The existing finance library is limited to compound interest and simple ordinary annuity problems. I thought it would be nice to expand it to handle all types of annuities as well as business investment decision methods so I created a new library called finance2. The names chosen for the functions are similar to those used in financial calculators, MS Excel, and Google Sheet so they are more intuitive.

The new library has new functions for:

- General annuities
- Annuity due
- Ordinary annuities
- Internal rate of return
- Net present value method
- Interest and principal portion of payments in amortization problems
- Days between dates
- I also developed and embedded a new TVM solver that can compute all types of annuity, unlike the existing one.

Please let me know if you have any questions.

### DIFF
--- a/assessment/libs/finance2.html
+++ b/assessment/libs/finance2.html
@@ -2,7 +2,7 @@
   <a name="top"></a>
   <h1>Macro Library finance2</h1>
 
-<p><span style="font-size: medium;">Financial Functions. Version 1. September 2020</span></p>
+<p><span style="font-size: medium;">Financial Functions. Version 1. October 2020</span></p>
 <p><span style="font-size: medium;">Author: Amir Tavangar</span></p>
 <p><span style="font-size: medium;">&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;Humber College, Toronto, ON </span>
 <br /><br />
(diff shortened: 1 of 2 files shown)


drlippman:
Thanks for the contribution!  I do have a couple questions/requests.

1) Please prefix the function names, like `finance_PV` or something similar.  This helps avoid potential name conflicts with other macro libraries and current or future built-in functions.

2)  It looks like your TVM solver is doing the computations in the backend (on the server).  The calculations for TVM are very simple, so if it'd be possible to rewrite that to do the calculations using Javascript, it would make the page a lot faster.   Also, would you be open to having the final product either included in the distribution or hosted elsewhere?  The current host seems pretty slow, and I like to try to avoid building in use of resources I can't guarantee will continue to exist in the future.

ahtavangar:
Thank you, David, for your response.

1. Absolutely. Can I use fin prefix (for finance), for instance, fin_PV, fin_PMT, etc? The shorter the better. Otherwise, I'll use finance_.

2. The TVM solver is currently on google servers and the reason it is slow might be that I have limited the resources it uses. The solver is written in Python (Flask). The thing is that I am pretty comfortable coding in python but don't have much experience coding in Javascript. Sure, I'd be happy to host it somewhere else, e.g., Amazon servers. It might resolve the slow calculations as well.

Please let me know.

Amir Tavangar

drlippman:
Yeah, fin_ is fine for a prefix.

If you're willing to share the code you're using for the TVM solver calculations, I could probably re-implement them in javascript so it won't need any back-end resources to operate.

drlippman:
It looks like there are some existing libraries for these calcualtions, like [tvm-finance.js](https://github.com/kgkars/tvm-financejs).  Maybe check out that library and see if it does the necessary calculations, and if so I can help adapt your front-end to use it if you don't want to tackle that yourself.

ahtavangar:
Hi David,

I have prefixed the name of the functions and updated both files. Also, I have given you access to my TVM solver repo. Please take a look at it and let me know how you'd like us to proceed with implementing the codes in java.

Thanks,
Amir

ahtavangar:
No, that library requires the user to calculate the interest per period first and then input it to the function. My solver gets the annual interest rate and does the calculation itself.

drlippman:
Thanks again for your work on this!  I'm going to go ahead and merge it, but I am going to trade out the TVM solver for now until we can speed up your new one.  I did extend the old one to handle different P/Y and C/Y, and to handle payment and the beginning.

PR #194, opened 2020-10-04
New Library for Complex Numbers

ahtavangar:
Hi David,

While creating complex questions, I realized there wasn't any library for complex numbers so I thought I'd create one. The functions in this library can be used for

- Converting from standard to polar form and vice versa
- Finding modulus, argument, and conjugate
- Multiplication, division, raising to a power, and taking the nth root
- Finding real or complex roots of quadratic equations

I have prefixed all the functions by cx_ . Please let me know what you think.

Amir


drlippman:
Thanks for your work on this.  A couple thoughts/questions.

1. Please try to create a new git branch for each new feature, so that the pull request only includes the new feature changes.  This PR currently includes your finance files too, which causes merge problems.

2. What is the intended use of functions like `cx_polDeg`?  I ask because the output format, using an HTML entity (which needs an ending semicolon, BTW) looks like it's intended for direct display, but wouldn't work input into the math typesetting or as an $answer to an ntuple question.

3. cx_polEu has an apparent typo in the third line, with a . instead of a * or space.

4. Since the system's default tolerance is .1% relative error, I'm a little concerned about the built-in rounding to 3 decimal places.  It'd be nice to at least allow that to be controllable, which you do have in some places but not in others like `cx_real`

5. Your `cx_mul`, `cx_pow`, etc. functions take different forms of input and seem to return strings.  My main concern is that this doesn't make it possible to chain together several complex operations, though I don't know how often one would need to do that.  

A possible alternate approach, if you wanted to change it, would be to have the input and output of all the functions be an array  `[real,imaginary]`, then have a separate function for formatting the final output.  That way someone could do something like:
```
$a = cx_mul([2,3],[4,-1])
$b = cx_mul($a, [3,1])
$disp = cx_format($b)
```
But that would require the author to write more code for single-operation use.

ahtavangar:
Thank you, David, for your feedback.

1. My bad. Sure, I'll create a new branch and a new pull-request after addressing the above points.
2. You are right. It was intended to be used in the n-tuple questions. I've removed the degree symbol. 
3. Yes it was a typo. I replaced the "." with space.
4. I'll add the rounding option to all functions where applicable.
5. cx_pow only raises one complex number to an exponent so the input would be easier to be just the real and imaginary parts as float. However, cx_mul always gets at least two complex numbers. As for output, I assumed that working with array output might not be as straightforward as a single object for question authors. That being said, I like your suggestion about making all the outputs the same so they can be used as input for the other functions and introducing a new function to display a formatted result.

Thanks again.
-Amir

ahtavangar:
Hi David, Quick question... Do you know if $showplot works in localhost? I am trying to test a function the shows the plot of a complex number, but the question doesn't show any grid even for a simple plot.

ahtavangar:
Thanks again for your suggestions. 

1. I changed the functions' inputs and outputs to be an array([a,b]) so they can be used in other functions as well. This allowed me to combine some of the functions, for example, all degree and radian functions are now combined.
2.  I also, added addition and subtraction functions, though it is easy even without a function, I thought I'd add them to have a complete library.
3. Finally, I have added a plot function, which I couldn't test on my local. I tried to simulate it on myOpenMath, and it worked. So I was hoping you could check if it works.

-Amir

drlippman:
showplot should work on localhost, yes.   I'll review these changes soon, thanks.

ahtavangar:
Hi David,

Not sure if you have gotten a chance to look at the new PR I sent for the
complex library, but I was wondering if the conflict has been resolved in
the new PR. Please let me know.

Amir

PR #201, opened 2020-10-26
User Email Addresses for Registraton and Login (Add support for $CFG['emailAsSID'] option)

xyztextbooks:
Changes in the Pull request enable the use of email addresses for registration and login. Detailed Development notes are here:
  https://docs.google.com/document/d/1fAhsDDIyJsvnlR79o8W9PcUa1ks4MD3QxC7AzPRbjJQ/edit?usp=sharing

 - When the new $CFG['emailAsSID'] option is enabled, the SID (username) field is hidden from users, so users register and login with their email address
 - Under the hood, the SID field is still used for registration and login, but it is set equal to the email field's value wherever possible (e.g., registration and login forms, and forms that allow for changing of user email addresses).
 - Requires database modifications to increase length of imas_users.SID field to 100
 - Requires modification of $loginprompt, $longloginprompt, and $loginformat vars in config.php
 - Complete usage notes appears in the new 'Using Email Addresses for Login' of readme.md

Files that were not modified (that are related to usernames):
 - batchcreateinstr.php
 - newinstructor-ipeds.php
 - modltidomaincred case in admin/actions.php

Several of the modified forms ("New Instructor Account Request", "Modify User Profile", "Admin -> New User", "Roster->Enroll a New Student") will allow an email address that is in another account's email field (as long as it is unique to the SID field). This will accommodate legacy situations that previously created multiple accounts with the same email.

### DIFF
--- a/actions.php
+++ b/actions.php
@@ -114,35 +114,42 @@ require_once("includes/sanitize.php");
 			$_POST['courseid'] = Sanitize::courseId(trim($_POST['courseselect']));
 			$_POST['ekey'] = '';
 		}
-		if (!isset($_GET['confirmed'])) {
-			//look for existing account. ignore any LTI accounts
-			$stm = $DBH->prepare("SELECT SID FROM imas_users WHERE email=:email AND SID NOT LIKE 'lti-%'");
-			$stm->execute(array(':email'=>$_POST['email']));
-			if ($stm->rowCount()>0) {
-				$nologo = true;
-				require("header.php");
-				echo '<form method="post" action="actions.php?action=newuser&amp;confirmed=true'.$gb.'">';
-				echo '<input type="hidden" name="SID" value="'.Sanitize::encodeStringForDisplay($_POST['SID']).'" />';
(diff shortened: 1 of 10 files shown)


drlippman:
I like the idea of this, but I worry about the confusion that would be caused by prompting for email but expecting a username from people who have legacy usernames.  Also, that editing a user's info would seem to change their username to their email without them possibly realizing.

I don't have an idea offhand of how to improve this, given that many systems already will have multiple user accounts with the same email.

xyztextbooks:
I agree that these changes are not ideal for legacy installations. Here at XYZ, we have three iMathAS deployments that already use email for login thanks to changes we made to our iMathAS fork. My main hope for this pull request was to get these changes into the iMathAS master branch and thereby avoid having to manually merge code every time we update iMathAS.

If you are still willing to entertain the idea of merging these changes, I suppose there are a few options:
 - document this option as "for new installations only"
 - document this option as "experimental" with full disclosure of the risks of using it
 - leave the changes undocumented and therefore usable only by those-in-the-know

drlippman:
Fair enough.  Being a config option, I suppose figuring out how to transition from the default to emails can wait.

There's a lot of code changes here, so it may take me a while to finish reviewing this.

PR #208, opened 2020-11-29
Prime

nick-chura:
I think this fixes the isprime(n) function so it works for n=1 through n=999999 -- an issue mentioned in the MOM forum.


nick-chura:
There is no good reason why this took 4 commits! I struggle with Git-efficiency.

drlippman:
Thanks, but I think I found a simpler fix.

PR #210, opened 2020-12-17
Dates Macro Library

SussexGooner:
I added the dates macro library that includes date manipulation functionality for question writing. Question 601187 contains test code for this library

The library includes the following functions:

1    dates_addDays($date,$days,$fmt): add days to date
2    dates_addWeeks($date,$weeks,$fmt): add weeks to date
3    dates_addMonths($date,$months,$fmt): add months to date
4    dates_addYears($date,$years,$fmt): add years to date
5    dates_EOMonth($date,$fmt): end of month
6    dates_BOMonth($date,$fmt): beginning of month
7    dates_diffDays($date1,$date2,$fmt): number of days between dates
8    dates_randdate($from,$to,$fmt): random date between dates
9    dates_dateFormat($date,$newFmt,$fmt): returns a date in a new format.  Allows for things like retrieving the day of the week, the name of the month etc.

thanks,

Daniel Brown

### DIFF
--- a/assessment/libs/dates.php
+++ b/assessment/libs/dates.php
@@ -4,102 +4,102 @@
 //          University of Colorado Boulder
 
 global $allowedmacros;
-array_push($allowedmacros,"dates_addDays","dates_addWeeks","dates_addMonths","dates_addYears","dates_EOMonth",
-"dates_BOMonth","dates_diffDays","dates_randdate","dates_dateFormat");
+array_push($allowedmacros,"dates_adddays","dates_addweeks","dates_addmonths","dates_addyears","dates_eomonth",
+"dates_bomonth","dates_diffdays","dates_randdate","dates_dateformat");
 
-// function addDays
+// function adddays
 // adds a number of days to an existing date
 // uses $fmt as format of date inputs and outputs
-function dates_addDays($date, $days,$fmt = "F j, Y"){
+function dates_adddays($date, $days,$fmt = "F j, Y"){
 	$datetime = date_create_from_Format($fmt,$date);
 	if ($datetime==false) {
-		echo "date_addDays::could not create date from string. The date is " . $date . " and the format is " . $fmt . ".";
+		echo "date_adddays::could not create date from string. The date is " . $date . " and the format is " . $fmt . ".";
 		return false;
 	}
 	$datetime->modify('+' . $days .' days');
 	return $datetime->format($fmt);
 }
 
-// function addWeeks
+// function addweeks
 // adds a number of weeks to an existing date
 // uses $fmt as format of date inputs and outputs
-function dates_addWeeks($date, $weeks,$fmt = "F j, Y"){
+function dates_addweeks($date, $weeks,$fmt = "F j, Y"){
 	$datetime = date_create_from_Format($fmt,$date);
 	if ($datetime==false) {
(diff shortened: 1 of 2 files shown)


assessment/libs/dates.php
```diff
@@ -0,0 +1,143 @@
...
+
+global $allowedmacros;
+array_push($allowedmacros,"dates_addDays","dates_addWeeks","dates_addMonths","dates_addYears","dates_EOMonth",
+"dates_BOMonth","dates_diffDays","dates_randdate","dates_dateFormat");
+
+// function addDays
+// adds a number of days to an existing date
+// uses $fmt as format of date inputs and outputs
+function dates_addDays($date, $days,$fmt = "F j, Y"){
+	$datetime = date_create_from_Format($fmt,$date);
```
drlippman, line 14: I think the function name is all lower case.  This applies everywhere this function is used.

assessment/libs/dates.php
```diff
@@ -0,0 +1,143 @@
...
+		return false;
+	}
+	$dt2 = date_create_from_format($fmt,$date2);
+	if ($dt2==false) {
+		echo "date_diffDays::could not create date from string. The date is " . $date2 . " and the format is " . $fmt . ".";
+		return false;
+	}
+
+	$interval = $dt1->diff($dt2);
+	return (int) $interval->format("%r%a");
```
drlippman, line 107: What happens if the difference is not a whole number of days?  Should this be rounded rather than cast to int?

drlippman (changes requested):
Generally looks good.

assessment/libs/dates.php
SussexGooner: Just to make sure, are you saying you want functions to be all lowercase rather than mixed upper and lower, or that I have mixed upper and lower incorrectly somewhere with dates_addDays?

thanks,
Daniel

assessment/libs/dates.php
SussexGooner: That's a great question and a use case I had not thought about.  As this is for financial applications like interest calculations, I will truncate first and then take the difference (and mention this in the documentation). One typically assumes that one receives a full day of interest irrespective of when during the day a deposit is made.

assessment/libs/dates.php
SussexGooner: apologies - I completely misread your comment.  I will change the function names to lower case.

assessment/libs/dates.php
SussexGooner: Hi,

I just tested the functionality and found that the existing implementation was as I wanted.  The difference between today at 3:10pm and tomorrow at 3:10am is 1 day and the difference between today at 3:10pm and tomorrow at 5:30pm are both one. I updated the test question to test this specific functionality.

SussexGooner:
Hi,
I'm new to github, so I am unsure whether I am following the correct process for you.  I have committed changes regarding the upper/lower case and tested/documented the behavior of diffdays.

Let me know if you have other comments and/or changes.  I really appreciate your help.

thanks,

Daniel

drlippman:
I think you misunderstood my comment.  I didn't have any issue with your functions having capitals in them.  My issue was the built-in function `date_create_from_format` that you are calling as `date_create_from_Format`, which the wrong capitalization.

SussexGooner:
As this is incorporated into the finderiv pull request, I am closing this pull request

thanks,

Daniel

PR #211, opened 2020-12-23
Feature.shapes

nick-chura:
This is the first version of the "shapes" macro library. It comes with an html help file.

The purpose is to simplify the creation of basic shapes with labels, using the existing showasciisvg macro in IMathAS.

### DIFF
--- a/assessment/libs/shapes.php
+++ b/assessment/libs/shapes.php
@@ -1,18 +1,19 @@
 <?php
 
 global $allowedmacros;
-array_push($allowedmacros,"draw_circle","draw_circlesector","draw_square","draw_rectangle","draw_triangle","draw_polygon");
+array_push($allowedmacros,"draw_circle","draw_circlesector","draw_square","draw_rectangle","draw_triangle","draw_polygon","draw_prismcubes");
 
 //--------------------------------------------draw_circle()----------------------------------------------------
-// circle("[center,[label]]","[radius,[label]]","[diameter]","angle,measurement,[label,point]")
+// circle("[center,[label]]","[radius,[label]]","[diameter,[label]]","[angle,measurement,[label]]")
 // Each "option" is a list in quotes. Here are the available options:
 // "center,[label]" This draws the center point with an optional label.
 // "radius,[label]" This draws a radius with an optional label.
 // "diameter,[label]" This draws a diameter with an optional label.
-// "angle,measurement,[label,point]" This draws an angle in standard position, with optional angle label and point on circle.
+// "angle,measurement,[label]" This draws an angle in standard position, with optional angle label.
 // Note: Angles are given in degrees, and degree symbol is displayed by default. To not display degree symbol, use e.g. "label rad".
 // "point,angle,[label]" This draws a point on the circle at "angle" degrees, counterclockwise from positive x-direction. Optional label on point.
-// Note: To get the "pi" symbol in an angle label, use e.g. "3 pi/4" for the label. This is for display only -- angle measurements must be given in degrees.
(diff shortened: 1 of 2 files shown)


drlippman:
Generally looks great!  A few questions:

1) in the basic angles and sides options for draw_triangle, is there a way to give it to always draw the triangle in the same orientation, like with right angle in the lower left corner?

2) for the "altitude", is there a way to label the altitude length itself?

3) Is there a way to control the size of the final diagram?  It seems to use the default 300x300.

As a bit of a random comment, I notice that often the resulting graph ends up having a lot of blank space.  I don't think that's a big deal, but it could make some diagrams harder to read, if the actual shape doesn't get the full space available.

nick-chura:
1. I added the "horizontal" option to have the longest side on the bottom horizontal, but that could be changed for right triangles. Are you thinking primarily of right triangles?
2. I considered adding a label for altitudes, but since that was the last option I wrote, I maybe just forgot. It would be easy to add, but I don't know how well I could optimize the label placement. I'll see what I can come up with.
3. Oh, yes. I kept the size at 300x300 since any smaller and the labels tended to interfere with each other. I can add an argument for "size,length,width".
4. Is the blank space thing specific to certain of the shapes?

drlippman:
1.  Yeah, mainly for right triangles.  For some reason it feels like it's common to see right triangles drawn with a horizontal leg.

4. The place I noticed it in particular was on the circle segment.  This really isn't a big deal.

nick-chura:
I've made some changes.

1. Triangles and polygons are now drawn in a "standard" position by default. [Vertices labeled counterclockwise from bottom-left, with bottom edge drawn horizontally.] Removed the "horizontal" option, added a "rotate" option for random rotation of the shape.
2. Added a "size" option to change the size of the image.
3. Fixed the image size for draw_circlesector.
4. Added a draw_prismcubes function.

assessment/libs/shapes.php
```diff
@@ -0,0 +1,1465 @@
...
+      $pointKey = $key;
+    }
+    if (in_array("axes",$in)) {
+      $args = "stroke='grey';line([-1.3,0],[1.3,0]);line([0,-1.3],[0,1.3]);stroke='black';";
+    }
+  }
+  
+  //number of sides
+  $n = $argsArray[0][0];
+  if (!is_numeric($n) || abs($n-round($n))>1E-9 || $n < 3) {
```
drlippman, line 1314: Small request: can you add a max value for $n too, to prevent the loops from getting too outrageous?  Maybe a max of 50 sides or something?

assessment/libs/shapes.php
```diff
@@ -0,0 +1,1465 @@
...
+  
+  for ($i=0;$i<($height+1);$i++) {
+    $args = $args . "line([0,$i],[$length,$i]);";
+    $args = $args . "line([$length,$i],[$length+$depth/sqrt(2),$i+$depth/sqrt(2)]);";
+  }
+  for ($i=0;$i<($length+1);$i++) {
+    $args = $args . "line([$i,0],[$i,$height]);";
+    $args = $args . "line([$i,$height],[$i+$depth/sqrt(2),$height+$depth/sqrt(2)]);";
+  }
+  for ($i=0;$i<($depth+1);$i++) {
```
drlippman, line 1450: These three loops seem like they have the potential to generate a LOT of lines, if height, length, or depth was large.  Could we add a max for these?

drlippman:
Looking good.  I added a couple comments on places where it'd be good to a max limit on the parameter to prevent excessive looping.

nick-chura:
I fixed those things. Let me know if there's anything else you'd like changed.

PR #212, opened 2020-12-27
finderiv macro library (and dates)

SussexGooner:
Hi,

Happy holidays. I have added a macro library of financial derivatives functionality. I decided to do this as a separate library rather than include it into finance or finance2 as the functionality seemed tangential to those libraries. Question 606302 is set up as a test harness for this library. 

I believe that the dates macro library is also included in this pull request. It is not completely necessary to release that library at the same time but I have used the dates macro library when testing this library.  I believe that this branch incorporates the changes you requested for the dates library.

Please let me know if you have comments or would like changes.

thanks,

Daniel Brown


assessment/libs/finderiv.php
```diff
@@ -0,0 +1,537 @@
...
+	if ($numTypes!=$numpossizes) {
+		echo 'finderiv_payout: type and position size arrays must be of the same size.';
+		return false;
+	}
+	if ($numTypes!=$numStrikes) {
+		echo 'finderiv_payout: type and strike arrays must be of the same size.';
+		return false;
+	}
+	$total = 0;
+    for ($i=0;$i<$numTypes;$i++) {
```
drlippman, line 44: Since you're looping on this, it'd be good to add a reasonable size check above (some max value it can't exceed), to prevent super-big $numTypes values that could cause system-locking loops.

assessment/libs/finderiv.php
```diff
@@ -0,0 +1,537 @@
...
+	if (!is_numeric($rf) ) {
+		echo "finderiv_fwdprice: invalid risk-free interest rate";
+		return false;
+	}
+	if (!is_numeric($yield) ) {
+		echo "finderiv_fwdprice: invalid yield";
+		return false;
+	}
+
+	$todaydate = date_create_from_Format($fmt,$today);
```
drlippman, line 89: This should be `date_create_from_format`, no cap on the Format.  Same everywhere else this function is used.

drlippman (changes requested):
Looks good.  Just a couple minor change requests.

assessment/libs/finderiv.php
SussexGooner: I added a maximum number.

assessment/libs/finderiv.php
SussexGooner: Sorry for the confusion.  I have corrected this.

PR #217, opened 2021-01-05
Anova added to stats lib

ahtavangar:
Thank you, David, for your feedback. I have made the changes and hopefully, it is all good. Not sure why I always have issues with github!


drlippman:
Merged manually.

PR #222, opened 2021-01-25
finderiv and dates: Bug fix and new functionality

SussexGooner:
I hope these changes are ok with you. Thank you for supporting this system. I have become a huge fan over the course of the last semester. Also, I am not sure if I need to do something, to allow the changes to be automatically merged. Let me know if there is more that I need to do.

This incorporates two sets of changes to dates and finderiv macro libraries.

The dates documentation had the dates_dateformat function miswritten.

The changes to finderiv include

- A bug fix to immdate and equityfutdate where odd months were not handled correctly and a set of parentheses were missing.
- Additions to the payout function to allow for an array of price levels instead of just one, allow for types to be specified more generally (for example call has to be in the string rather than the string has to be call), and a position with a bad type is ignored rather than an error being given.
- A new function checkpayout that compares a payout to a set of lines returned by a student drawn graph. This simplifies question writing and allows for the somewhat complex functionality to be hidden from the question writer.

thanks,

Daniel Brown


drlippman:
Merged manually, since this PR had merge conflicts with master.

I also rewrote a couple of your changes to use array_walk to handle the case differences, rather than using preg_match.

PR #226, opened 2021-02-15
Matrixformat fraction

ahtavangar:
Hi David,

I have added matrixformatfrac(), matrixscalarfrac(), and matrixrowscalefrac(). However, only matrixformatfrac() would suffice to display fractional entries. For example, the output of matrixscalar() and matrixrowscale() can be formatted by matrixformatfrac() to display fractions.

Please let me know what you think and whether we should keep or remove matrixscalarfrac() and matrixrowscalefrac().

Also, if there are any other macros for matrices you would like me to work on, please let me know.

Amir


drlippman:
While I'm not a big fan of using decimaltofraction so much, I'm OK with the `matrixformatfrac` function.  

The other two seem unnecessary as implemented, since they output the fractions as strings, but only work on inputs as numbers.  If those were implemented like matrixsolvefrac, which parses fractions and does actual fraction calculations rather than using decimaltofraction, that might be a useful separate function.

For now, I'm going to go ahead and just bring in the matrixformatfrac function.

PR #229, opened 2021-02-17
Matrix fraction2

ahtavangar:
Hi David,

Thanks for the feedback. This time, I have implemented fractionparse and other fraction operations for matrixformat, matrixscalar, rowscale, and rowcombine.
I have also replaced decimaltofraction() in the original matrixformat with fractionparse and fractionreduced.

Amir


drlippman:
Thanks for this.  After doing some testing, it seems that using decimaltofraction on the format pass is much more efficient than doing the running fraction calculations along the way.  This is probably partly due to a rewrite of decimaltofraction a while back that made it much more efficient.  And for any reasonable calculations its should be accurate enough, so I'm going to opt to just stick with the simpler matrixformatfrac.

ahtavangar:
Yes, I would keep it too.

PR #231, opened 2021-02-22
New macro: to construct matrix given eigenvalues/vectors

ahtavangar:
Hi David,

As you suggested, I have created a macro to construct the original matrix given eigenvalues and eigenvectors. The macro accepts both real and complex entries.

Amir


drlippman:
Nice!  A few thoughts, which you're welcome to use or ignore:

1) It seems your $iscomplex==true branch can also handle real eigenvalues/vectors, so would there be any downside in dropping the $iscomplex parameter, and just always using that code branch?  That way the user wouldn't have to explicitly declare it's complex.

2) Could the $size be inferred by the length of the first eigenvector, then checked against the lengths of the others to catch inconsistencies?  That way the user wouldn't have to declare the size.

3) I think you could skip the work of the `matrixdet` call if you replace the `matrixinverse` call with `matrixsolve($p, matrixidentity($size), true)`.  That  third parameters is the "silent error" parameter, and will cause it to silently return `false` if the matrix isn't invertible.  So then if it returns false, you can give your linearly independent warning.

4) I just noticed there's an existing `matrixFormMatrixFromEigValEigVec` function which essentially does what the real number version of your function does.  I wonder if it'd make sense to have your function replace this one?  Though the eigenvector format is different I think; the existing function takes the P matrix as the input for the eigenvectors.

Unrelated to that, I'm guessing the resulting matrix often has ugly decimals because of the inverse calculation.  Probably not avoidable.  I know when I've written questions involving eigenvalues I'll often use `matrixrandunreduce` to generate eigenvectors so the P^-1 matrix will have nice integer values.  I wonder if there's a way to generate complex eigenvectors in a similar way, so that the resulting matrix will have integer values...

ahtavangar:
Hi David,

Thanks for the feedback. I have made the changes.

1. I created the real part and complex part separately and then combined them into one function so I hadn't realized that the complex part would take care of both types. Thanks.
2. Yes, I removed $size from the arguments.
3. Matrix inverse and set were replaced by `matrixsolve`.
4. Yes, I noticed it too but since there wasn't any information about the format of the arguments I couldn't get it to work properly. Now that you mentioned the input is P, then it works with the transpose of $vec. I think `matrixFormMatrixFromEigValEigVec` can be replaced with the new function.
5. I did a preliminary experiment with the row operation on P of the complex eigenvectors and it seems P^-1 would keep integer values. Maybe I can write a function to get complex $val and $vec and some integer n (random row combination) and return a random matrix with or without new eigenvectors/eigenvalues.

drlippman:
Looks great.  Do you want to finish your explorations with the complex eigenvector stuff first, or should I go ahead and merge as-is?

BTW, somehow this eigen branch has your finance changes, which makes it impossible for me to automatically merge the branch.  Not a big deal, but it would make my life easier if the branch only contained the new changes.

drlippman:
Here's my stab at a function to generate random eigenvectors (real and complex) based on a set of eigenvalues so that the eigenvectors are nice integers and the resulting matrix will have nice integers.

```
function matrixFriendlyEigenVecsFromVals($values) {
    $size = 0;
    foreach ($values as $v) {
        $size += is_array($v) ? 2 : 1;
    }
    $m = matrixidentity($size);
        $mult = nonzerodiffrands(-3,3,6);
        for ($i=0; $i<6; $i++) {
        list($sr,$er) = diffrands(0,$size-2,2); // leave last row alone for now
        $m = matrixrowcombine($m,$sr,$mult[$i],$er,1,$er);
    }
    $mult = nonzerodiffrands(-3,3,2);
    for ($i=0; $i<2; $i++) {
        $sr = rand(0,$size-2);
        $m = matrixrowcombine($m,$sr,$mult[$i],$size-1,1,$size-1);
    }
    $mult = rands(1,3,$size);
    $n = 0;
    // real evals will get one row as evec
    // complex evals will get one row for imag, and next for real part
    foreach ($values as $v) {
        if (is_array($v)) {
            // add multiple of last row to real rows
            if ($n+1 != $size-1) {
                $m = matrixrowcombine($m,$size-1,$mult[$i],$n+1,1,$n+1);
            }
            $n += 2;
        } else {
            $n++;
        }
    }
    
    $out = [];
    $r = 0;
    foreach ($values as $v) {
        if (!is_array($v)) { // real: use row as e-vec
            $out[] = $m[$r];
            $r++;
        } else { // complex: use two rows as real and complex
            $vec = [];
            for ($i=0; $i<$size; $i++) {
                $vec[] = [$m[$r+1][$i], $m[$r][$i]];
            }
            $out[] = $vec;
            $r += 2;
        }
    }
    return $out;
}
```
That may be overkill, not work totally right, or produce ugly matrices; I haven't tested it completely.  If something like this was used, it'd be nice to have a function also that converted the eigenvectors into a more human-friendly format too.

drlippman:
I've been thinking about this more, and am trying to consider what use cases there will be for these functions.  And it's got me wondering whether I should have been more open to your original proposal of a function to find eigenvals/vecs of 2x2.  I'm thinking there are probably two primary use cases of eigen functions:

1) Doing 2x2 problems, where starting with the matrix might be more important in cases like application problems, and where finding the eigenvals/vecs is not too complex.  Maybe it would be nice to have functions for this case.

2) Doing larger problems, where finding eigenvals/vecs is hard, and where it would be nice for them to have friendly values.  For this use case, constructing a matrix based on the eigeninfo probably makes more sense.  But then it got me thinking: since random eigenvecs lead to ugly matrices, I wonder if it'd be better for this function to just take an array of eigenvalues and to return both a set of randomly generated eigenvectors and the final resulting matrix.  Basically a combo of the function I posted earlier and your work (which could be combined in a more efficient way, I'd imagine).

Are there other use cases I'm not considering?  Is it still worth having the function to generate the matrix given the eigenvectors, or is it safe to assume that people would rather have integer eigenvectors that lead to integer matrices?

ahtavangar:
> Looks great. Do you want to finish your explorations with the complex eigenvector stuff first, or should I go ahead and merge as-is?
> 
> BTW, somehow this eigen branch has your finance changes, which makes it impossible for me to automatically merge the branch. Not a big deal, but it would make my life easier if the branch only contained the new changes.

Maybe let us explore the other stuff first, and then merge the finalized work.
At this point, I don't really know why there are still some conflicts in the new branches I create. Maybe because once I developed on my master and now whatever I do to undo that just doesn't work. I'll try again.

ahtavangar:
> I've been thinking about this more, and am trying to consider what use cases there will be for these functions. And it's got me wondering whether I should have been more open to your original proposal of a function to find eigenvals/vecs of 2x2. I'm thinking there are probably two primary use cases of eigen functions:
> 
> 1. Doing 2x2 problems, where starting with the matrix might be more important in cases like application problems, and where finding the eigenvals/vecs is not too complex.  Maybe it would be nice to have functions for this case.
> 2. Doing larger problems, where finding eigenvals/vecs is hard, and where it would be nice for them to have friendly values.  For this use case, constructing a matrix based on the eigeninfo probably makes more sense.  But then it got me thinking: since random eigenvecs lead to ugly matrices, I wonder if it'd be better for this function to just take an array of eigenvalues and to return both a set of randomly generated eigenvectors and the final resulting matrix.  Basically a combo of the function I posted earlier and your work (which could be combined in a more efficient way, I'd imagine).
> 
> Are there other use cases I'm not considering? Is it still worth having the function to generate the matrix given the eigenvectors, or is it safe to assume that people would rather have integer eigenvectors that lead to integer matrices?

1. The applications I can think of are in differential equations. Another is in diagonalization and evaluating the powers of matrices.  Others would be just finding eigenvalues and eigenvectors in linear algebra (up to 3x3). It'd be nice to have a function that can return random 2x2 matrices (as well as eigeninfo) that have integer eigenvalues (real and complex).
2. Good idea. For larger matrices, we can take nice eigenvalues and return the matrix as well as eigninfo. I'll work on it.

drlippman:
I took a stab at this in #234 .  See what you think.

drlippman:
For now I'm going to go with the other version; we can come back to this one if there seems to be need in the future.

PR #235, opened 2021-02-26
version 15

Cash4Programming:
version 15 - updated american length language, update pi symbol to pi, update verify equal sign to =, ~~, or HTML entity, verifypi added


Cash4Programming:
Updated to version 16

PR #244, opened 2021-03-15
cx_matrixreduce added to complex library

ahtavangar:
Hi David,

I have added cx_matrixreduce to the complex library. It can handle matrices with both real and complex entries.

Amir


ahtavangar:
I'll resolve the conflict and send another pull.

PR #245, opened 2021-03-17
Simplex update with debugging code

Cash4Programming:
This is a debugging release so I can see where in the code the NAN is occurring.


Cash4Programming:
New version

PR #246, opened 2021-03-17
added to the help file for comparefunctions

nick-chura:
I noticed the "integers" option for comparefunctions() wasn't in the help file.


drlippman:
Thanks!

PR #255, opened 2021-04-15
added comparentuples function

nick-chura:
You suggested having a comparentuples() function could be helpful. Here is something I put together for both ntuple and calcntuple.

Format:  comparentuples(a, b, [tol], [option])

Compares ntuples a and b of same dimension. Mimics comparenumbers() in that you can set single tol for all entries in the tuples, or can use an array or list of tol's to apply on a by-element basis. [Can also use the "|tol" format for absolute tolerances.]

By default, comparentuples() will also compare the style of brackets surrounding the ntuple (returning false if they are mismatched). Can add an optional argument "ignoreparens" that will only compare the entries in the ntuple.


drlippman:
Looks good.  One thought:  Do we want to check the entry is actually an ntuple? Particularly in the case of the ignoreparen argument.  Though I guess it's pretty unlikely someone would enter something that is not a proper ntuple but still matches the values after stripping off the first and last values.

nick-chura:
I added a check to make sure the ntuples are enclosed by ( ), [ ], { } or < >.

PR #275, opened 2021-08-13
support for sq, cu and per

nick-chura:
Supports common exponent language for units:

"sq unit" => "unit unit"
"square unit" => "unit unit"
"unit squared" => "unit unit"

"cu unit" => "unit unit unit"
"cubic unit" => "unit unit unit"
"unit cubed" => "unit unit unit"

"foo per bar" => "foo / bar"

I decided against things like "ft sq" since it would make "ft sq cm" ambiguous.

I added "sq, square, squared" etc to the JS regex of approved units, but I still get errors such as: unknown unit "square"


drlippman:
I'm not getting the error you're describing with the javascript - it seems ot be working OK to me.

However, because of the way the units check is set up it checks each unit separately, so "3 cm square" and "3 cm sq" don't trigger a unit warning, but are rejected in scoring.  The only workaround for that would be to implement your rewrite in the javascript before the units check, but that'd be kind of messy.

One other thing, the loop on lines 365-371 of units.php looks a bit expensive to me, since it's running 4 regex for every unit in the list.  It'd be nice to wrap it in something that checks to see if there's even a need to run any of those, maybe something like `if (preg_match('/(sq\b|square|cu\b|cubed|cubic)/', $unitsExpression))`.  That way we skip the huge loop if there's not even a square or cube to resolve.

Oh, and you did leave an echo at line 373 of units.php :)

nick-chura:
I've run into this before, where changing the AMhelpers2.js file doesn't update what I'm seeing. [Maybe my system is caching that file?] It seems to right itself after a period, but the lapse makes testing difficult.

I see what you mean about it not throwing an error with "3 cm square" but also marking it incorrect. I guess I could add to the processNumber() function in AMhelpers2.js (and the _min files) in order to strip out "sq, cu, per" etc under the right conditions before it checks against the list of units.

Good point about the huge loop -- I'll address that.

drlippman:
Setting `$CFG['assess2-use-vue-dev'] = true;` in your config.php will help (since that cause it to use AMhelpers2.js instead of assess2_min.js), and hitting Control-F5 in your browser will often force a reload of cached resources.

nick-chura:
I've made changes so that "1 ft cube" throws an error message, but "1 ft cubed" does not. Please test and tell me what you think.

PR #277, opened 2021-08-19
new approach for handling sq, cu, per, etc

nick-chura:
Here is a better(?) approach for interpreting sq, cu, per, etc. It doesn't have any large loops of searches like the previous approach.

Now:
"1 sq ft" or "1 square ft" => "1 ft*ft"
"1 ft cubed" => "1 ft*ft*ft"
"1 ft per sec" => "1 ft / sec"

It also interprets some non-standard but still reasonable things:
"1 sq ft^3" => "1 ft^3 * 1 ft^3"
"1 ft^2 cubed" => "1 ft^2 * ft^2 * ft^2"


drlippman:
Out of curiosity, is there a reason you decided to support "sq ft^3" and similar?  Is there a more standard unit in the same format you were trying to accomodate with that?

drlippman:
Also, your comments mention `Change "ft sq" to "ft*ft"`, but the code doesn't actually do that particular transform (it does "sq ft" and "ft squared" but not "ft sq").  Was that intentional?  

Perhaps it was to remove the ambiguity from something like "ft sq L" which could arguably be "ft^2 L" or "ft L^2"

nick-chura:
_is there a reason you decided to support "sq ft^3" and similar?_

I assume that to many students, "sq ft" and "ft^2" may feel like fundamental units, so it seemed reasonable (and mathematically sound) to extend the meaning of "sq blah" and "blah sq" to allow ft^2 in place of "blah". [However, I drew the line at things like "sq cu ft".] I figured if we're allowing sq, cu, etc., then in the off chance a student enters "sq ft^3", it interprets it...correctly?

_Also, your comments mention Change "ft sq" to "ft*ft", but the code doesn't actually do that particular transform (it does "sq ft" and "ft squared" but not "ft sq"). Was that intentional?_

That's definitely a typo in that comment in the code. It should read:

Change "**sq ft**" to "ft*ft"

I intentionally made _sq, square, cu and cubic_ the only prefixes that double or triple a unit, and _squared and cubed_ the only suffixes.

So, _1 cm cubed per sq sec_ should be fine, but _1 cm cu per sec sq_ should throw two errors.

nick-chura:
_Perhaps it was to remove the ambiguity from something like "ft sq L" which could arguably be "ft^2 L" or "ft L^2"_

Exactly.

drlippman:
I guess I'm just wondering whether there's actually a use case for "sq ft^3".  Like, would that ever actually come up in a problem?  If not, is it worth coding special handling for something that never comes up?

nick-chura:
Perhaps not, though I wasn't thinking whether it would come up in a problem, but rather if a student enters it, should the system try to interpret it. I did check that Wolfram|Alpha interprets things like ft^2 cubed as ft^6, though it wants to call sq ft cubed the same as ft^8.

Let me know if you'd like me to remove it.

PR #288, opened 2021-10-06
Minor update to track down code warning

Cash4Programming:
Minor update to start tracking down code errors.  If I need to update any function name - email me so I an fix them.


Cash4Programming:
Fixed bug in simplexsolve2()

PR #289, opened 2021-10-10
Added a comparenumberswithunits function

nick-chura:
Here's a comparenumberswithunits() function.

I thought this could be helpful in conditional-type problems where partial credit is to be earned for writing an equivalent measurement with units. Like the other compare...() functions, this allows for a relative or absolute tolerance.

When used, it calls the libs/units.php file, so it can be used whether or not the answerformat includes "units".

Do you think this could be useful?


nick-chura:
Based on a question/comment in a MOM forum, I had considered also adding an option to ignore the numerical value and just check for compatible units (e.g. comparing "3 feet per second" and "2000 miles per hour" would evaluate to true, since they are both velocities), but I thought to first wait and see if you thought having a compare function even sounded like a good idea. [It was suggested in the forum that it would be nice to be able to give partial credit if the number was wrong but the units were "correct". This is one way to make that a possiblity.]

PR #299, opened 2021-11-20
two-sample_t_test_added

ahtavangar:
Hi David,

I thought I'd add the two-sample t-test to the stat library. It computes t-test for the dependent and independent samples with equal and unequal variances and sizes.

Amir


drlippman:
Merging manually.  I messed something up with github.

PR #318, opened 2022-02-17
Add.unit prefix plural case

nick-chura:
Changes here include:

- Allowing metric prefixes only on SI units, SI derived units, and some others that have used them historically (e.g. kilofeet)
- Allowing abbreviated metric prefixes on abbreviated SI units, abbreviated SI derived units, and a few others (e.g. kft)
- Allowing units to be pluralized with an 's' where it made sense (e.g. meters is allowed, but gausss is not)
- Allowing case insensitivity, except when it introduces ambiguity (e.g. Meters is okay, but FT is not since it could be interpreted as 'feet' or 'femptotesla')

Please check this out and let me know if I need to address anything.


nick-chura:
I realize I didn't give a rationale for choosing what could be prefixed, etc. Looking online at a variety of resources, I landed on the following choices:
1. Abbreviated units, such as "m" or "Pa, should not be pluralized with an ending "s". [I do see this a fair bit with U.S. customary units such as "qts" or "pts", but I left those nonpluralizable.]
2. Abbreviated metric prefixes should only be allowed on abbreviated metric units, e.g. "1.21 GW" (with some exceptions like "ft" and "Btu" that still have some common use). [Note that we already had separate entries for "g" and "mg" and "kg", etc. which was not the best use of space.] Spelled out metric units can use long metric prefixes (e.g. "1.21 gigaWatts"). 
3. I wanted to allow some flexibility with case-sensitivity where it made sense, so all spelled out units are case insensitive and abbreviated units are case sensitive. Together with allowing plurals, this allows "newton", "Newton", "newtons", "Newtons", "NEWTONS", etc. without having to include each of those in the main units array. [We already had some of these listed separately, so I wanted to keep and expand this to other units while decreasing the size of the units array.]

Also, I wasn't sure if defining the javascript regexes should have been done in the units.php file and then just called by the javascript files, or if you wouldn't mind those all living in the AMhelpers2.js file.

drlippman:
Note to self: look to check that British/Canadian spellings of litre, metre, deca, gramme are handled

PR #324, opened 2022-04-12
Initialize  biology.php

bibbca:
Create biology.php, Create biology.html.

Add 2 Functions: bio_randcodon(), bio_getcodonname()


bibbca:
Adding a few new features to the macro library. I'll reopen this pull request when I'm finished with them.

PR #330, opened 2022-05-03
Typo in poly3

Cash4Programming:
Missing a $ on line 165. Adding it causes the text to be the expected variable.


drlippman:
It looks like you didn't actually commit the fix, but I'll go ahead and add the $ myself :)

Cash4Programming:
Darn it. I haven't used the Git features for so long I forgot how to do it.
Thanks for the fix.

Mike

PR #331, opened 2022-05-04
Add new unit: Molarity

bibbca:
Add Molarity (mol/L) as a unit, also add M to regex.


drlippman:
Thanks for this.  I have a feeling this is going to conflict with #318 which I've been slow to merge and review.  Perhaps @nick-chura can weigh-in on whether the M molarity unit is going to cause problems in the revised units scheme?

nick-chura:
Adding M as a unit doesn't interfere with #318 -- it just would need to be added to javascript/assess2.js, javascript/AMhelpers2.js and the .min files.

Just two thoughts on adding it:

1. [Wolfram|Alpha](https://www.wolframalpha.com/input?i=convert+1+mol%2FL+to+M) and [Google](https://www.google.com/search?client=firefox-b-1-d&q=convert+1+mol%2FL+to+M) both define 1 M as 1 mol/L, so there would be some conflict in our defining it as 1 mol/m^3. Wikipedia [agrees](https://en.wikipedia.org/wiki/Molar_concentration) that 1 mol/m^3 is the SI definition, though says it is less popular than 1 mol/L in chemistry literature.
2. A personal fear is that adding M (by either definition) may cause a lot of reasonable answers to be marked wrong, such as a student wanting to answer "5 meters" but typing "5 M" and not getting an answerbox warning message (like they do now).

So, I think it's a reasonable add, though I have some reservations because the system doesn't (yet) have a way to tell a student how it is interpreting a unit such as "m" or "M".

nick-chura:
Also, in the format of #318, I think M would look like

'M' => [1,array(0,-3,0,0,0,0,0,1,0,0)**,1,0,0**], // Molarity

because it would allow only abbreviated metric prefixes, we don't want to allow it to be pluralized with an ending 's', and it is case sensitive.

drlippman:
Yeah, I had the same concern about a student accidentally entering "5 M" when they mean "5 m", but there will be cases with units where we'll run into that.

Wikipedia does suggest that 1 mol/L is commonly designated 1 M, while it seems to suggest mol/m^3 is designed c.  And indeed it later says that 1 mol/m^3 = 10^-3 M.  Given that, it sounds like the definition of M should instead be

`'M' => [0.001,array(0,-3,0,0,0,0,0,1,0,0),1,0,0], // Molarity`

Caleb, as the science teacher, does that seem right to you?

drlippman:
Wait, I think I have that backwards.  1km = 1000m gets coded as `'km' => [1000,array(0,1,0,0,0,0,0,0,0,0)]`, 

Then since 1 mol/m^3 = 10^-3 M, we'd have 1 M = 10^3 mol/m^3, which would be 
`'M' => [1000,array(0,-3,0,0,0,0,0,1,0,0),1,0,0], // Molarity`

1 M = 10^3 mol/m^3

bibbca:
Honestly, I wrote that line copying the line for Liters.  Since in my mind, 1 M = 1 mol / L.  
`'L' => [0.001,array(array(0,3,0,0,0,0,0,0,0,0)],`

So, long story short, that does seem right to me.

drlippman:
Great. I need to finish reviewing Nick's contribution, which will conflict with this one.  Once I get that one done, we can figure out how to merge this change in.  Hopefully next week.

bibbca:
Thanks so much!

PR #332, opened 2022-05-06
Updated the Conversion spelling

Cash4Programming:
Updated the spelling of deca/deka based on the detected browser language.


Cash4Programming:
I hope I did this correctly this time. If not message me in WAMAP so I can fix this.  Thanks.

Mike

drlippman:
Why?  The gettext `_` function already allows localization, and handles it in general for any language.  Why restrict it to only handle a few languages?  (Though I suppose most languages probably don't spell the units differently)

I'm also concerned with displaying the British spelling for the units here, when the units handling code for number questions only handles the US spelling.

Cash4Programming:
Updated the function to use the gettext() _ function from a function at the top of the file (conversion).

Added simplexcreatelatexinequalities in the simplex and updated the simplex html

drlippman:
Your original code had already used _ for localization, so I'm still confused what problem this change is trying to solve.

drlippman:
Also, I just realized that gettext is based on server-declared language, so can't adapt to the client's language preference like you were trying to do here.  If your goal is for the same server to deliver both spellings, it might be necessary to go back to your original model, but with gettext used on the fallback case to account for other languages.

Cash4Programming:
I had a request that deca be spelled deka as that was the way it was done
in Ben's book. I was going to change it and realized that I would need to
update it all over the file.

I responded that it could be spelled either way. I did update the file do
it would be easier to maintain.

Cash4Programming:
I want the server to have the same units as the named units.

I was hoping to show the user a localized version.

Cash4Programming:
I will close this and request and use the gettext() as the fallback.

PR #333, opened 2022-05-07
Allow for superscript charges

bibbca:
Adds regex to allow for superscript charges in HTML display chemistry library macro of the form digit+, digit-, +, or -.


drlippman:
Please be sure to test all cases.  Your regex doesn't quite work here, since you have two paren capture groups, so the `(\+|\-)` group would be $2, and doesn't result in the desired output.

I think you can fix it by changing the regex so it has just one capture group that contains the 4 possibilities:
`$c = preg_replace('/\^(\d\+|\d\-|\+|\-)/','<sup>$1</sup>',$c);`

Is it possible the charges could ever be two-digit?  If so, then you'd want:
`$c = preg_replace('/\^(\d+\+|\d+\-|\+|\-)/','<sup>$1</sup>',$c);`

bibbca:
Thanks so much! I'm really new to regex. This is very helpful

PR #334, opened 2022-05-07
Updated

Cash4Programming:
I have updated the conversion to use gettext as the default.
Updated simple to include create simplexcreatelatexinequalities
Updated simple.html to include the new functions and show DEPRECIATED ones.


drlippman:
I'm embarrassed to say it, but I think that if your intent was to adjust the spelling of units for different users on the same server based on their browser's stated language preference, your original approach of using the language header was probably the right one, though we'd still want to use gettext on the fallback to handle servers configured for non-English languages.

I'm not sure if you can easily revert back to that version.  If not, I can try to find your earlier version.

Cash4Programming:
No problems - I copied an earlier version from the github commits and updated the files.

My logic is for right now test for en-GB, then en-US, if neither match then use the gettext() to return the word.

ingodahn:
Can I help with de?

drlippman:
@ingodahn once we get this merged, if you want to updates the de translations, that'd be great.

Cash4Programming:
That would be awesome!

Mike

Cash4Programming:
Yes, please!

drlippman:
BTW Mike, it'd be much appreciated in the future if when developing a new feature you could do it in a branch, then PR that branch.  That helps isolate those changes, so multiple changes aren't mixed together.

drlippman:
I ended up making some changes, using a mix of your techniques.  It uses the browser-stated English variant to use the GB/CA spelling when asked for, then falls back to using gettext for translation otherwise (which default to en-US if no translation language is specified)

PR #335, opened 2022-05-12
Conversion update

Cash4Programming:
Updated the conversion file with German language option.


drlippman:
This doesn't seem to contain any changes. There is no need to do something special in the conversion script for German; those translations are handled by gettext.

Cash4Programming:
It's the wrong file.

PR #336, opened 2022-05-12
Correct File

Cash4Programming:
This should be the correct file.


drlippman:
I do _not_ want to be coding in translations for individual languages like this.  I don't want to try to adapt the language based on the browser-declared language, especially not on a file-by-file basis.  gettext is what we use to handle internationalization, so there's no need to do this; if the server is set to use de as the language, the gettext translation will be used for that language, so there's no need to code translations into the script.

The only reason I was OK the en-GB/en-CA being handled separately is because they were 99% the same language, so probably wouldn't need/want a full gettext translation, plus we have some Canadian users on MOM, so being able to make a small adjustment for a few words was nice.

Cash4Programming:
OK - I will eliminate the version I have and rollback to the one in your
master repository.

PR #340, opened 2022-05-19
Revised Update to Logic Answerformat and New Setexp Answerformat

rubberdrop:
May 19, 2022 by Avery Flowers

Changes to 'logic' answerformat:
- Added 'xor' (oplus) functionality for exclusive or
- Changed ~ for the more commonly used \neg

New 'setexp' answerformat:
- Added 'cap/cup/^c/-/ominus' functionality (intersection, union, complement, set difference, symmetric difference)
- Created new comparesetexp function to compare student answer to correct answer (similar to comparelogic)
- Updated help.html with these changes

New 'logic' library

New 'setexp' library

KNOWN ISSUES:
- Unable to have 'xor' display correctly in question text.
- Unable to have 'ominus' display correctly in question text, and issues with it displaying in mathquill.
I am unable to determine the source of these issues. Truly sorry!


drlippman:
Your issue with 'xor' and 'ominus' in the question text is because we use MathJax's implementation of Asciimath for rendering, which don't include those symbols, and there's not a great way to extend it with new symbols.  

How essential is the symmetric difference operator, or for that matter the xor operator?  I've only taught basic set theory, so never have had need for either.

ingodahn:
Xor is an essential operation in circuit design. It can be used as a 
single basis to define all other propositional operators.

rubberdrop:
They're fairly common symbols in the first-year undergraduate courses I've
taught and attended, but if it can't be done easily then perhaps it's not
too big of an issue. Personally, I could live without it if needed. The two
libraries I submitted ("logic" and "setexp") both have functions to enable
the display of these operations in question text (but unfortunately
"setexp" does this by literally replacing 'ominus' with it's unicode
symbol). So there are workarounds...

drlippman:
This has been merged.  I made a few revisions while merging, including fixing the randomization functions in logic.php and setexp.php to use the repeatable randomizer functions.

Thanks for the contribution!

PR #342, opened 2022-05-24
Conversion.function2

Cash4Programming:
Changed all string references to two arrays $unit and $unitabbr. Converted conversion functions to return an array of values with the first value an array of conversion strings (equivalent to the original function) added checks in conversionArea and conversionVolume for switched sign and tick arguments and corrected them (should not be needed)


Cash4Programming:
See updated request #343

PR #343, opened 2022-05-27
Conversion

Cash4Programming:
version 24:
Removed:
conversion_detectlanguage functions that were eliminated as they interferred with gettext function from version 23.

Added:
changed all string references to two arrays $unit and $unitabbr. Converted conversion functions to return an array of values with the first value an array of conversion strings (equivalent to the original function) through added checks in conversionAreaand conversionVolume for switched sign and tick arguments and corrected them (should not be needed)


Cash4Programming:
David,

Any concerns about this updated version?

Mike

PR #356, opened 2022-10-04
edits to help file jsxgraph

nick-chura:
I noticed that setting the 'face' of points, gliders and intersections won't accept camelCase, as in the help file, but rather only all lowercase. I edited the help file in those places.


drlippman:
That's interesting.  The JSXgraph documentation says it should be camelcase, but the code itself indeed expects lowercase.  Weird.

PR #367, opened 2022-12-10
Bug fixes and division by zero checks

Cash4Programming:
Bug fixes and division by zero checks


Cash4Programming:
Bug fixes and division by zero checks. I added the is_numeric at the beginning of the routine so I only needed to check it once rather then in each loop.

PR #368, opened 2022-12-12
Bug Fix

Cash4Programming:
Bug fix for Division by zero on line 3177


assessment/libs/simplex.php
```diff
@@ -3362,7 +3367,7 @@ function simplexdisplaytable() {
                         // R1C(Last)
                         if($mode>0) { $Tableau.= "<td $nopad>&nbsp;</td><td $nopad>&nbsp;</td>\r\n";} // add augemented column
                     }
-                    if(!empty($headers[$cloop])) {
+                    if((!is_null($headers[$cloop]))&&($headers[$cloop]!="")) {
```
drlippman, line 3370: In the future, can you not revert this change?  The !is_null check isn't enough to prevent errors in the case where $headers[$cloop] is undefined.

PR #369, opened 2022-12-13
Fixed logic bug in simplexreadsolutionarray

Cash4Programming:
Fixed logic bug in simplexreadsolutionarray


Cash4Programming:
pulled so I don't revert a change.  I think I have fixed it on my end...

PR #374, opened 2022-12-28
makes stuansready catch empty drawing in conditional

nick-chura:
I noticed that stuansready was returning true for drawing parts submitted blank in conditional problems. Of course, feel free to deny/close this if there's a better way to fix it.


drlippman:
Your approach here is fine, but since ';;;;;;;;' would be a really unlikely answer to any other question, I think I'm going to just add a check for that, even if $anstypes isn't provided.  That might help some existing questions without needing changes.

PR #398, opened 2023-06-09
Conversion and Displaytable updates

Cash4Programming:
updated both the Conversion and the MatrixDisplayTable


drlippman:
This is showing a conflict with master.  Can you please resolve then update the PR?

Cash4Programming:
Should be resolved.

drlippman:
After merging, I'm going to change `$headers[$cloop]!=null` to `isset($headers[$cloop])`, as the latter won't throw an error if the entry is undefined.

PR #401, opened 2023-07-14
create parseStuansAsMatrix()

rpruim:
As per the discussion [in this forum](https://www.myopenmath.com/forums/posts.php?view=0&cid=1&page=1&forum=374413&thread=1347456&r=64b1712993c82), I've created a function that takes a student-supplied calculated matrix and returns it *as a matrix* rather than as an array.

I have not updated the documentation.  Is that done by manually updating the HTML?  If so, we could put something like this in there:

```html
<h3><a name="parseStuansAsMatrix">parseStuansAsMatrix</a></h3>
<p>parseStuansAsMatrix(student answer)</p>
<p>Converts a students matrix answer into a matrix</p>
<p>student answer can come from <code>$stuanswers[$thisq]</code> for single-part
questions, or <code>getstuans($stuanswers, $thisq, (part index))</code> for
multi-part</p>
```

### DIFF
--- a/assessment/libs/matrix.php
+++ b/assessment/libs/matrix.php
@@ -16,7 +16,7 @@ array_push($allowedmacros,"matrix","matrixformat","matrixformatfrac","matrixsyst
 	"matrixIsRowsLinInd","matrixIsColsLinInd","matrixIsEigVec","matrixIsEigVal",
 	"matrixGetRowSpace","matrixGetColumnSpace","matrixFromEigenvals","matrixFormatEigenvecs",
 	"matrixAxbHasSolution","matrixAspansB","matrixAbasisForB",
-	"matrixGetMinor","matrixDet","matrixRandomMatrix","matrixParseStuans");
+	"matrixGetMinor","matrixDet","matrixRandomMatrix","matrixParseStuans","parseStuansAsMatrix");
 
 //matrix(vals,rows,cols)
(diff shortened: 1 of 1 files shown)


rpruim:
I've added an additional argument to control whether symbolic expressions are evaluated.

drlippman:
Thanks for this.  However since the evaluated values for the matrix are already present in $stuanswersval, I'm hesitant to have this function do the work of evaluating the expressions a second time, as that's pretty inefficient.

I'm thinking a more efficient pattern would be to have the optional second argument be the $stuanswersval entry.  If provided, the function could still use $stuanswers from the first argument to determine the dimension, then use the second entry for the values.

It probably should also be noted that the function only works for matrix questions without $answersize set.  That definitely has the potential to be confusing.  Perhaps it would be better to have the function inputs be `parseStuansAsMatrix(student answer, [answersize, student answer val])`, where answersize would be null if answersize wasn't set on the question.

rpruim:
This snuck past me in July.  Sorry for that.

I'm basically in support of some good way to get a matrix and know it's shape without the question writer needing to do hacky things each time this happens.  It is a bit confusing that parseStuAnsAsMatrix() doesn't parse *as a matrix* but only provides an array.

drlippman:
Cleaning up old PRs. Closing as matrixParseStuans was alternately implemented.

PR #402, opened 2023-07-31
Allow the Canvas OIDC hostname to be set in config.

kittysnacks:
This is to support Canvas' new OIDC auth endpoints for LTI 1.3.

Use the following variables in config to set hostnames:
- `$CFG['LTI']['canvas/oidc/prod'] = 'sso.canvaslms.com';`
- `$CFG['LTI']['canvas/oidc/beta'] = 'sso.beta.canvaslms.com';`
- `$CFG['LTI']['canvas/oidc/test'] = 'sso.test.canvaslms.com';`

If the above variables are not set, default to the original
hostnames:
- canvas.instructure.com
- canvas.beta.instructure.com
- canvas.test.instructure.com

### DIFF
--- a/lti/Database.php
+++ b/lti/Database.php
@@ -136,8 +136,8 @@ class Imathas_LTI_Database implements LTI\Database
         if ($row === false || $row === null) {
             if (!empty($GLOBALS['CFG']['LTI']['autoreg']) && trim($client_id) != '') {
                 $canvas_oidc_hostname = $GLOBALS['CFG']['LTI']['canvas/oidc/prod'] ?? 'canvas.instructure.com';
-                $canvas_beta_oidc_hostname = $GLOBALS['CFG']['LTI']['canvas/oidc/prod'] ?? 'canvas.beta.instructure.com';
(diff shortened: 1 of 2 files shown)


drlippman:
Out of curiosity, is there a reason to have this as something that needs to be set in $CFG, instead of just changing the values?  Also, it looks like this doesn't do anything to changing the URL used by existing registrations, right?

kittysnacks:
> Out of curiosity, is there a reason to have this as something that needs to be set in $CFG, instead of just changing the values? Also, it looks like this doesn't do anything to changing the URL used by existing registrations, right?

I was asked to make the change in a manner that would move the values into config, vs values inlined with code. This does not modify any existing registrations, so a script will need to be run against the DB to update those.

drlippman:
Gotcha.  Given that Canvas is recommending all tools change to use the new URLs, I'm not going to merge this PR, and am just change the URLs hardcoded in the system in the near future.  At that time I'll also add a migration to rewrite any existing registrations.

PR #408, opened 2023-09-20
Bux fixes

Cash4Programming:
Fixed identical sub-expressions bug.


drlippman:
Will update manually to avoid un-doing previous changes

PR #411, opened 2023-10-24
Update conversion.php

Cash4Programming:
Fixed errors in conversion factors for US Capacity


Cash4Programming:
Just notified of this error today - fixing typos.

PR #415, opened 2024-01-06
undefined array key error with byelement and partweights

nick-chura:
I noticed the following error when using "byelement" and "partweights" in an n-tuple problem:

_Caught warning in the question code: Undefined array key 1 on line 285 in file /var/app/current/assess2/questions/scorepart/NTupleScorePart.php_


drlippman:
Thanks, but the changes you made don't seem to fix the issue.  Line 285 really needs to change to `if (!empty($matchedparts[$pwi]))`.  I'll make that change.

PR #432, opened 2025-01-08
set question preview popup location to same window as active browser

sparrowme:
Account for multi-display desktop environments for placing the position of question preview popup window.


drlippman:
I know I asked you to update the PR with cross-browser compatibility, but I'm going to go ahead and merge then add that myself.

PR #442, opened 2025-06-13
Added one-time blank-parts warning with bypass

mlandrum22:
Description:
Show a “Warning” dialog when a student submits with any parts blank, but only on the first click—subsequent submits go through.
	•	Store: add blankOverride flag in basicstore.js
	•	Guard: on first partial submit set errorMsg='blankparts' + blankOverride=true; on next submit clear it and proceed
	•	Reset: clear blankOverride when all parts are filled, a new question loads, or a full submit succeeds
	•	UI: update en.json to label the dialog “Warning” and adjust the message; tweak ErrorDialog.vue to use that title

How to test:
1. Load a multipart question.
2. Submit with some blanks → see the warning once.
3. Submit again → it submits.
4. Click **Get a similar question** to load a new question.
5. Repeat steps 2–3 on the regenerated question to verify the warning appears again.

### DIFF
--- a/assess2/vue-src/src/basicstore.js
+++ b/assess2/vue-src/src/basicstore.js
@@ -13,6 +13,7 @@ export const store = reactive({
   inTransit: false,
   autoSaving: false,
   errorMsg: null,
+  blankOverride: false,
   confirmObj: null,
   lastLoaded: [],
   inProgress: false,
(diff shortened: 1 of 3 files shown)


drlippman:
Since in most cases the system will only grade submitted parts, I know a lot of instructors encourage students to submit only one part at a time, see if they're correct, then continue to the next part. In those situations, showing a warning whenever there is a blank part might be really annoying.

Perhaps this feature could be limited to cases when all parts are submitted together, or when scores/results aren't shown during the assessment?

There are, of course, also questions where parts are intentionally left blank, like questions with multiple-choice parts where some parts have values to be filled in if that part is selected.  While only showing the warning once then allowing them to submit anyway should be ok, I do worry some students might get concerned that they didn't answer the question correctly in cases like that.

drlippman:
Also, some question parts might `jsparams['submitblank']` set, and in those cases the blank should be submitted without prompting this warning.

PR #456, opened 2026-06-27
Fix for loop to support decreasing ranges (e.g. for($i=5..1))

DaleWMyers:
for($i=5..1) previously did nothing because the loop was hardcoded to always increment. This fix detects when start > end and uses a decrementing loop instead.


drlippman:
Thanks for the suggestion and contribution.  I ended up rewriting it a bit to a disallowed variable for the endpoints to prevent  tampering, and took the opportunity to rewrite the non-integer endpoint handling to something more sensible.

PR #464, opened 2026-08-01
Add Ctrl-H find-and-replace to CodeMirror question editors

DaleWMyers:
Binds Ctrl-H to the replace command in all four CodeMirror editors (controlEditor, setupQtextEditor, setupRandvarEditor, setupKeepcodeEditor). Also adds the missing CodeMirror search/dialog addon files (searchcursor.js, search.js, dialog.js, dialog.css) which were not in the compressed bundle, and fixes solver.js which was clobbering controlEditor extraKeys on init.

Note, if the cursor is outside of Common Control or Question Text, Chrome, Firefox, and Edge will send you to their search history. But, if the cursor is in Common Control or Question Text, you will get a two-step find and replace dialogue.

### DIFF
--- a/forms.php
+++ b/forms.php
@@ -555,7 +555,7 @@ switch($_GET['action']) {
 
 			function bta (o) {
 				let pre = "=?BINARY?B?", suf = "?=";
-				for (let k in o) {
+				for (let k of Object.keys(o)) {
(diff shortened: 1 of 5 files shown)


drlippman:
Thanks for this. I won't be able to review this for a few weeks.

PR #471, opened 2026-09-15
Add in-place fraction calculator library

DaleWMyers:
The file "inplacecalculator.php" includes the PHP function "inplacecalculator()". This generates a one-line fraction calculator with an input box, a button, and an output box. Enter a fraction to be calculated or reduced, e.g.,  "1/2+1/3". Then either click the button or press <ctrl-enter>. The reduced fraction result, "5/6", appears in the output box. 
   The calculations are done with the JavaScript function "reduce()" in "inplacecalculator.js". This file also includes "ratEquiv(,)" that tests if two formulas are algebraically equivalent. It is modeled on MOM's "comparefunctions(,)". 
   For this button, Common Control needs loadlibrary("inplacecalculator") and $calculator = inplacecalculator(). In Question text, place $calculator where the calculator should appear. 

The file "inplacecalculator.js" provides answerbox functionality. If you have "x+1/2+1/3" in an answerbox, and highlight the subformula "1/2+1/3", then clicking a calculator button or pressing <ctrl-enter> will calculate and replace the selected text, giving "x+5/6". No replacement occurs in any of the following cases. In particular, replacements that change an answer value are not allowed.
   Case 1. The selection contains anything other than digits, +,-,*,/,(,),[,]. "^" is allowed only for the powers -2, -1, 0, 1, 2. 
   Case 2. The selected segment is undefined. E.g., no replacement occurs with the selection "1/(1-1)".
   Case 3. The proposed replacement changes the formula's value. No replacement occurs if "ratEquiv(,)" shows that the replacement changes the value of the parent formula. E.g., no replacement occurs if "2+1/3" is selected in "x+1/2+1/3".
   Case 4. In-place replacement requires MathQuill-rendered answer boxes; with MathQuill display off, nothing happens.

The file "inplacecalculator.php" also includes the PHP function "inplacecalc()". This generates a small button labeled "#" with no input or output box. When clicked, it provides the same answerbox functionality as does pressing <ctrl-enter> or clicking the larger button. To replace the default label "#" with say, "%", write inplacecalc("%") instead of inplacecalc().

In some "edge" cases, the replacement has to be prefaced with "+" or "*". If "-1+2" is selected in "x-1+2", replacing "-1+2" with "1" gives "x1". In this "edge" case, "+" is added before the replacement to get "x+1". These edge cases are detailed in QID 1905848

The file "inplacecalculator.html" is a helper file that describes the inplacecalculator and includes an example of how to use it.

SIDE EFFECTS.
It adds one global, window.InPlaceCalculator. No other globals; no existing names touched. reduce(), ratEquiv() and preview() aren't globals — they're window.InPlaceCalculator.reduce, .ratEquiv and .preview.

There are three delegated listeners on document: mousedown and click (which ignore everything except the calculator buttons) and keydown for ctrl-enter. The keydown handler calls preventDefault only when a replacement was actually made, so when the calculator has nothing to do, the keystroke passes through untouched.

The script injects its stylesheet into document.head as <style id="inplacecalculator-style">, once. Putting it there rather than beside the buttons means the styling survives a question redrawing its text after a submit.

The script tag is written with every button rather than once per page, for the same reason: a redrawn question must get it back. A second execution does nothing.

On the page it uses the classes ratIn, ratOut, ratGo, ratGo2 and inplacebutton.

Nothing loads on pages that don't call inplacecalculator() or inplacecalc().

It requires BigInt, so no IE11. Arithmetic is exact — no rounding, no overflow, at any size.
 
MOM QUESTIONS EXAMPLES AND TESTS. 
These questions and tests should work after the PR is merged. Before merge, they will show "inplacecalculator.js is not loaded".
A simple MOM question with answer type "multipart".  QID 1903421 
A simple MOM question with answer type "Algebraic Expression". QID 1903423 
A MOM question with many examples of in-place calculations.  QID 1903422 
A test suite for the JavaScript function "reduce()", which handles fractional calculations.  QID 1903887
A test suite for in-place calculation. It details the "edge" cases that depend on the context of the selected text.  QID 1905848   
           
This link, [inplacecalculator.test.js](https://math.hawaii.edu/~dale/inplacecalculator.test.js.txt), runs 203 tests under Node (node inplacecalculator.test.js), covering the arithmetic, the rewrite rules, and the button and Ctrl-Enter behavior. It needs jsdom, which stands in for the browser and for MathQuill. I ran it on Windows with Node 24.


DaleWMyers:
When I run this, I get the error message
   Error loading library inplacecalculator Eeek.. unallowed macro inplacecalculator
   QID 1903423
   Common Control
   loadlibrary("inplacecalculator")   
$calculator=inplacecalculator()   //also fails with $calculator = inplacecalc()
$answer="3"

   Question text
$calculator
$answerbox answer = 3

